PR #94, opened 2021-06-16
Extract toaster component

doctorn:
~~This isn't ready to be merged yet but I thought it would probably worth showing you guys what I'm up to.~~

The aim of all of these changes is to extract the toaster from the app model so that it lives as a stand-alone component. There's some more macro magic involved, but the result is pretty nice! You can use the toaster from the app like this:
```rust
self.toaster.send(Toast::error(format!("{}", error)));
```
And to add a new variety of toast you just stick another line in this [macro call](https://github.com/doctorn/homotopy-rs/blob/85c6502e9af2d99c6e9e11157adce94b38f035fb/homotopy-web/src/components/toast.rs#L42).

~~The most contentious part of all of this is the introduction of the `WeakComponentLink` type taken from [this example](https://github.com/yewstack/yew/tree/master/examples/nested_list). It's a bit of a hack, but for low traffic components (like the toaster), it lets you send messages directly to a child, rather than having to go all the way up to the top level. Let me know what you think!~~

I've spent a lot of time reading the Yew documentation and `Agent`s look like the best way to go for this sort of thing. The cool thing about this approach is that _any_ component can get access to the toaster just by calling `ToastAgent::dispatcher`.

I think we'll end up with a similar architecture for the settings implementation and I might try and port some more components to this style of communication.


jamievicary:
Can you give me instructions to clone this pull request locally and run it please.

NickHu:
@jamievicary I've updated the CI to generate build artifacts, so you can easily try out the versions at different commits. If you click the [Checks](https://github.com/homotopy-io/homotopy-rs/pull/94/checks) tab above, and then you'll see there's a 1 next to a label **Artifacts**. There's now an artifact called `web` which contains a zip file of the built app.

jamievicary:
Ah that's great. How do I run the artefact version. I've unzipped it and tried loading the index.html in a web browser (tried Chrome and Firefox) but I just get a blank page.

NickHu:
That's because web browsers disable a bunch of things when they operate from local files. See https://developer.mozilla.org/en-US/docs/Learn/Common_questions/set_up_a_local_testing_server#the_problem_with_testing_local_files.

The solution is to run a web server serving the unzipped directory, and then point your web browser to http://localhost at the appropriate port. If you have python installed it has a built in web server, which can be started with `python3 -m http.server`, or [mongoose](https://mongoose.ws/) which is really small.

jamievicary:
Great thanks. Does GitHub not automatically host everything anyway, e.g.
via GitHub pages? If the artefact was just committed unzipped wouldn't that
allow us to easily run the committed version.

jamievicary:
OK, I managed to play with it -- beautiful thanks Nathan!!!

zrho (approved):

PR #98, opened 2021-06-21
Add settings subsystem

doctorn:
This adds a new system for creating settings contexts and interacting with them. This currently doesn't support local storage, but I envisage local storage management as a separate system that can be hooked into various settings contexts.

To create a new settings context:

```rust
declare_settings! {
    pub struct MySettings {
        example_toggle: bool,
    }
}
```

Any component can then call

```rust
let mut settings = MySettings::connect(/* callback */); // This should be stored in component structs
settings.subscribe(&[MySettingsKey::example_toggle]);
```
to connect to the settings provider and subscribe to settings.

To perform an update:
```rust
self.settings.set_example_toggle(true);
```

See example in `src/app/settings.rs` for how this is connected up to a component.

(Blocked on #94)

@zrho @NickHu


jamievicary:
This looks excellent Nathan. @zrho please can you give your opinion.

zrho (approved):

PR #101, opened 2021-06-24
Extract sidebar from `App`

doctorn:
This definitely wants a review... The hope is that this PR gives us a foundation for implementing things like reconfigurable key-chords and generally makes adding UI features simpler (for the mean time I'm switching onto the graphics pipeline).

Some potentially contentious points:
- Rename `Proof` to `ProofState` 
- Alias `Proof` to `NodeData<Snapshot>` (implementing `Deref<Target = ProofState>`)
- Refactor history tree to avoid `RefCell`s
- Add `IdxVec` to `homotopy-core` (@Akvile1206 is using this anyway but worth pointing out)

@zrho @NickHu


homotopy-web/src/app/signature.rs
```diff
@@ -94,9 +93,9 @@ impl Component for SignatureView {
         // TODO: Folders/groups
         // TODO: On mobile, drag to the side to delete
         html! {
-            <Drawer title="Signature" class="signature">
+            <>
```
NickHu, line 96: Can we just remove this?

homotopy-web/src/app/project.rs
```diff
@@ -33,12 +32,12 @@ pub fn project_view(props: &Props) -> Html {
         }
     }));
     html! {
-        <Drawer title="Project" class="project">
+        <>
```
NickHu, line 35: Necessary?

homotopy-web/src/components/rawhtml.rs
```diff
@@ -4,33 +4,36 @@ use web_sys::Element;
...
+            props,
+            node_ref: NodeRef::default(),
+        }
     }
 
     fn update(&mut self, _: Self::Message) -> ShouldRender {
         true
     }
 
     fn change(&mut self, props: Self::Properties) -> ShouldRender {
```
NickHu, line 31: There's something in `yewtil` which is essentially this function. I don't remember if we're using it in the codebase, but if we are then please use it here too (otherwise it's probably not worthwhile).

NickHu (approved):
Looks good to me. I added a minor nitpicks to check.

homotopy-web/src/components/rawhtml.rs
doctorn: I don't think we're using it at the moment but it may become necessary when I start working on the WebGL stuff

doctorn: I'll have a look when I start authoring the render pipeline and switch over then if necessary

homotopy-web/src/app/project.rs
doctorn: I think this one actually is needed because there are a few top elements here but I totally agree for the signature one

PR #124, opened 2021-07-21
Add extensible data to rewrites

doctorn:


NickHu:
Let me just explain the motivations for this a little bit. Currently, rewrites encode zigzag maps sparsely, and in doing so they forget some information. For example, (by design) two distinct zigzag maps which are similar but have different codomains (e.g. one is a superset of the other) correspond to the same rewrite. I believe this occurs in degeneracy maps in normalisation, which needs to keep track of trivial identity maps which are forgotten by the standard encoding. Another thing we want is to encode (in a lightweight manner) *cubical* zigzag maps. These things can be achieved by tagging rewrites with some payload data, which @doctorn noticed can be neatly achieved while respecting hashconsing by making rewrites generic in their allocator. With this we can have a high-level unified description of 'rewrite-plus-extra-data'.

calintat:
This is great, thanks @doctorn !

BTW there is currently no way to access a rewrite's `payload`. You might want to add a function
```rust
impl<A> GenericRewriteN<A>
where
    A: RewriteAllocator,
{ 
    pub fn payload(&self) -> &A::Payload {
        &self.0.payload
    }
}
```

doctorn:
Totally forgot about that! Thanks!

PR #131, opened 2021-07-27
Add a variety of panzoom improvements

doctorn:
- Zoom control in top right (and reset zoom button)
- 0-dimensional diagrams
- Reset panzoom when switching diagrams


jamievicary:
Nathan this is gorgeous 🥇

jamievicary:
The dot for a 0-dimensional diagram is a bit too small, could we make it 3 times the size please?

PR #139, opened 2021-07-30
Signatures 2.0

NickHu:
image

* Generators are now individually rendered in a `GeneratorView` component
* Signature is now a tree rather than a hashmap (this is a **breaking
  change** regarding serialisation)
* The signature now supports folders: generators can be grouped into
  folders, which themselves may also contain nesting folders.
  Furthermore, these elements may be reordered via drag and drop

This one is fairly big, could do with some review.

### Todo

- [x] Folder renaming
- [x] Check compatibility on browsers other than Firefox desktop
- [x] Check that deletion of generators still recursively propagates correctly
- [x] Check that serialisation still behaves deterministically (at the moment `Tree`s are being traversed by their underlying `IdxVec`, which I think is probably not injective - many different `IdxVec`s will correspond to the same `Tree`, but maybe this doesn't actually happen while using the tool. Penny for your thoughts @doctorn?)
- [x] It looks a bit ugly


jamievicary:
Thanks for this Nick, this feature will be really excellent.

Unfortunately I think it's now too late to make breaking changes to
serialisation. Old-style serialisations will need to be accepted and
converted to a depth-1 tree.

If this requires iterating a version id in the serialisation format that
would be worthwhile.

homotopy-web/src/util.rs
```diff
@@ -0,0 +1 @@
+pub mod tree;
```
doctorn, line 1: In my rendering branch I've actually moved `tree.rs` into `homotopy-core`. Do you think it's better placed here or there?

homotopy-web/src/model/proof.rs
```diff
@@ -560,7 +612,11 @@ impl ProofState {
             let diagram: DiagramN = workspace.diagram.clone().try_into().unwrap();
             let generator: DiagramN = self
                 .signature
-                .get(&option.generator)
+                .iter()
+                .find_map(|item| match item {
+                    SignatureItem::Item((g, i)) if g == &option.generator => Some(i),
+                    _ => None,
```
doctorn, line 618: I think maybe we should inline the `Node` in the generator data so that we don't need to do this iteration everywhere.

homotopy-core/src/idx.rs
```diff
@@ -74,6 +75,17 @@ where
         self.raw.len()
     }
 
+    #[inline]
+    pub fn map<F, U>(self, mut f: F) -> IdxVec<I, U>
```
doctorn, line 79: We should have an `IdxMap` iterator returned here.

doctorn: (Just because this shouldn't require an allocation.)

homotopy-web/src/model/proof.rs
NickHu: I think this would make the serialisation non-deterministic, and make it depend on the order in which a workspace is created. Ideally we'd like to keep it deterministic.

doctorn:
Ok, so this is rebased along with a whole pile of changes. I'm still a long way from finishing it though. I'm hoping to get a chance to fight through some more this weekend.

jamievicary:
This is fantastic Nathan. Thanks for doing all this I can see it's quite a
huge job.

Is there some way we can change the way we're working to avoid these issues
in the future?

doctorn:
Still to-do tomorrow:
- [ ] Folder renaming
- [ ] Folder deletion
- [ ] Fix colour picker
- [ ] Check recursive deletion
- [ ] (Seperately fix panzoom crash)

Then we're good to merge! So close!

jamievicary:
Can't wait Nathan!!

jamievicary:
Hooray!

PR #144, opened 2021-08-05
Pipe errors through to interface

doctorn:
@NickHu unfortunately I don't think this is as useful we hoped... However, if someone who actually understands categories goes into `homotopy-core` and adds a bit more to the error types there, then I think this change will help a lot.

### DIFF
--- a/homotopy-core/src/contraction.rs
+++ b/homotopy-core/src/contraction.rs
@@ -62,7 +62,7 @@ pub enum ContractionError {
     Invalid,
     #[error("contraction ambiguous")]
     Ambiguous,
-    #[error("contraction fails to typecheck")]
+    #[error("contraction fails to typecheck: {0}")]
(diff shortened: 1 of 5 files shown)


jamievicary:
Ideally when a failure case is reached the core we would be apply to supply a string that propagates up and is displayed as a toast. We can then put all sorts of useful information in this string (i.e. which failure case of the contraction base case was triggered, etc.)

doctorn:
> Ideally when a failure case is reached the core we would be apply to supply a string that propagates up and is displayed as a toast. We can then put all sorts of useful information in this string (i.e. which failure case of the contraction base case was triggered, etc.)

So this change enables that, it's just the strings we're propogating are (at the moment) a little bit rubbish.

jamievicary:
That's completely fine, there was no motivation until now to add interesting information to those strings.

PR #150, opened 2021-08-18
Check if factorisation result is well-formed

calintat:


doctorn:
There's no reason not to just merge this is there?

doctorn (approved):

PR #156, opened 2021-08-24
Add well-formed check in constructors

calintat:
This will check that diagrams/rewrites are well-formed whenever a new one is constructed. This could be very slow, so it is only enabled in debug mode.


jamievicary:
Brilliant. How do we enable debug mode, could it be a global option in the settings drawer?

calintat:
You have to build the website in debug mode. Currently, beta.homotopy.io is built in release mode but I think we should build in debug mode. But it might be worth making this a global setting, so you can turn it on/off whenever you want. @doctorn can you help with this since you wrote the settings agent?

jamievicary:
Yes global setting please we will need to be turning this on and off at
runtime.

If global setting is impossible then it needs to be on for the live
version, so that Hao can use it by default.

Maybe it won't actually be so terribly slow due to all the caching.

zrho:
Having the online deployment be compiled in debug mode might be a problem since this turns off a couple optimizations which makes everything else much slower (I've seen orders of magnitude slowdown sometimes).

jamievicary:
We definitely need to be able to run the live beta version in a mode where
diagram correctness is constantly verified. This is so that Hao can use it
to expose any remaining bugs. It's not enough to know that Hao has produced
a malformed diagram, we need to know exactly how he did it, and that's only
possible with runtime correctness checkers at every interaction.

zrho:
In this case we could just do the checks even in release mode for now, until we have a toggle or are reasonably confident?

PR #157, opened 2021-08-25
Add error information to well-formed checker

calintat:


calintat:
The return type of the well-formed check was changed from `bool` to `Result<(), MalformedDiagram>` for diagrams and `Result<(), MalformedRewrite>` for rewrites, where `MalformedDiagram` and `MalformedRewrite` are some error types that specify why the diagram/rewrite fails the check.

jamievicary:
Great, is it easy to propagate this info as a toast? Would that be useful?

zrho:
It'd probably be best if that info showed up in the console instead of the toast so that one can look at it in peace.

calintat:
I agree with Lukas. This is for debugging purposes only.

PR #164, opened 2021-09-04
Make factorisation into an iterator

### LINKED ISSUE #160: Homotopy between 3d singularity and 2d singularity
**Demo file**
[homotopy between singularities of different dimensions.zip](https://github.com/homotopy-io/homotopy-rs/files/7070678/homotopy.between.singularities.of.different.dimensions.zip)

The following homotopy between a 3d singularity and a 2d singularity is viewed as 'ill-formed' in the current homotopy.io version. This is potentially caused because homotopy.io treats this move in a wrong way; instead of a homotopy between a 3d singularity and a 2d singularity, this should really be a homotopy between that 3d singularity and the identity of that 2d sigularity.

Screenshot 2021-08-28 at 13 16 47

calintat:
Fixes #160

### DIFF
--- a/homotopy-core/src/factorization.rs
+++ b/homotopy-core/src/factorization.rs
@@ -1,107 +1,128 @@
 use std::ops::Range;
 
-use crate::monotone::MonotoneIterator;
-use crate::Rewrite;
-use crate::{Diagram, Rewrite0};
-use crate::{Height, RewriteN};
-use thiserror::Error;
+use crate::monotone::{Monotone, MonotoneIterator};
+use crate::Height;
+use crate::{Diagram, DiagramN};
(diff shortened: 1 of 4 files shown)


homotopy-core/src/factorization.rs
```diff
@@ -1,107 +1,128 @@
...
+                            factorize(
+                                self.f.slice(si),
+                                self.g.slice(ti),
+                                self.source.slice(Height::Singular(si)).unwrap(),
+                                self.target.slice(Height::Singular(ti)).unwrap(),
+                            )
+                        })
+                        .multi_cartesian_product();
+                    self.cur = Some((h_mono, product));
+                    self.next()
```
doctorn, line 104: This would probably tidy up really nicely with a few more `?`s involved

PR #170, opened 2021-09-10
Add placeholder 3D & 4D diagram views

doctorn:


doctorn:
@calintat this should be good to merge when the CI passes

PR #177, opened 2021-09-17
Add 4D lighting

doctorn:


jamievicary:
Hi Nathan, is the linear idea working?

PR #183, opened 2021-09-24
Refactor graphs in `homotopy_core`

calintat:
`GraphBuilder` is now generic in the `RewriteAllocator` so it can also explode cubical graphs in cubicalisation.

I also simplified the keys since they are always coordinates.

### DIFF
--- a/homotopy-common/src/graph.rs
+++ b/homotopy-common/src/graph.rs
@@ -4,9 +4,6 @@ use crate::{declare_idx, idx::IdxVec};
 
 declare_idx! {
     pub struct Node = usize;
-}
-
-declare_idx! {
     pub struct Edge = usize;
 }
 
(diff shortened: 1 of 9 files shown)


calintat:
It is now generic in the `RewriteAllocator` so it can also explode cubical graphs in cubicalisation.

NickHu:
> I also simplified the keys since they are always coordinates.

I feel like there was a reason why it's generic, but I can't remember right now.

calintat:
> > I also simplified the keys since they are always coordinates.
> 
> I feel like there was a reason why it's generic, but I can't remember right now.

Think about it this way: if a node has coordinate `[a, b, c]`, its key will be `f(a, f(b, f(c)))` so you could always compute the key from the coordinate, so we don't lose any generality by getting rid of the key and replacing it with a coordinate.

doctorn (commented):
Looks really good!

PR #211, opened 2021-10-20
Tidy up diagram constructors

calintat:


calintat:
I refactored the diagram constructors, so we now have three constructors:
- `DiagramN::new(source, cospans)` which constructs a diagram directly from a source and list of cospans and does a consistency check unless that feature is turned off (this was previously called `new_unsafe`),
- `DiagramN::new_unsafe(source, cospans)` which is the same as `new` but does not do a consistency check, and
- `DiagramN::from_generator(generator, source, target)` which constructs a diagram from a source, target, and generator, and checks that the dimensions match and the globularity conditions hold (this was previously called `new`).

We use the safe constructors (i.e. `new` and `from_generator`) pretty much everywhere, except:
- in `DiagramN::rewrite_forward/backward` because these are called by `DiagramN::check_well_formed` so we previously were checking every slice twice which could be slow. Also, the result of a forward/backward rewrite is well-formed iff the original diagram was well-formed, so the check is redundant.
- in deserialization to ensure that it doesn't panics if we import a malformed proof when the consistency checks are turned on (instead, we manually check all the diagrams in the signature and workspace are well-formed to prevent the user from importing malformed proofs).

homotopy-core/src/diagram.rs
```diff
@@ -218,6 +219,14 @@ impl DiagramN {
         diagram
     }
 
+    /// Unsafe version of `new` which does not check if the diagram is well-formed.
+    pub(crate) fn new_unsafe(source: Diagram, cospans: Vec<Cospan>) -> Self {
```
NickHu, line 223: Maybe add `#[inline]` here

homotopy-core/src/diagram.rs
```diff
@@ -177,7 +177,11 @@ thread_local! {
 }
 
 impl DiagramN {
-    pub fn new<S, T>(generator: Generator, source: S, target: T) -> Result<Self, NewDiagramError>
+    pub fn from_generator<S, T>(
```
NickHu, line 180: Maybe this should be a `impl<S: Into<Diagram>, T: Into<Diagram>> TryFrom<(Generator, S, T)> for DiagramN`?

Issue #189, opened 2021-09-28
Mesh incorrectly built for pants unit


jamievicary:
The issue is most visible for low subdivision level.

[filename_todo (48).zip](https://github.com/homotopy-io/homotopy-rs/files/7243858/filename_todo.48.zip)
![homotopy io - Google Chrome 28_09_2021 13_14_06](https://user-images.githubusercontent.com/10602270/135084880-48cff90a-9a75-4ad2-bbe0-dd96322a0377.png)

doctorn:
I've spent an hour or so looking at this this evening and I still have absolutely no idea what the problem is. I'm going to work on mesh splitting as a priority and then come back to this when I've had a think about how to go about debugging it properly.

jamievicary:
Thanks for looking at it Nathan.

jamievicary:
Nathan can you explain again what this issue is. Is it the problem of, for a given cubical 3d mesh embedded in 4d, extracting the triangles that appear in any timeslice?

doctorn:
I wish I knew... I think it has to do with mesh extraction (as in some cubes that aren't being extracted from the cubical graph), but it could also be a typo anywhere in the rendering pipeline.

calintat:
@doctorn what's the smallest example where this happens?

doctorn:
I think lips 2 is a good start!
[lips_2.hom.zip](https://github.com/homotopy-io/homotopy-rs/files/7850495/lips_2.hom.zip)

calintat:
Interesting, I tried removing the boundary from the mesh to make it easier to debug and now it looks fine. Which is very bizarre, since the boundary only contains non-degenerate cubes.

calintat:
But the pants unit is still broken so I might do that one instead since it's a smaller mesh.

calintat:
Since this is two separate issues, I am going to close this in favour of #339 and #349.

PR #274, opened 2021-12-06
Add explicit simplicial geometry

doctorn:


jamievicary:
Thanks Nathan, what is this about?

doctorn:
This is a big refactor that separates simplicialisation from cubical layout / subdivision. It's a step on the path to the general 'inflation' procedure that takes (n - k) simplicies in an n-simplicial mesh and multiplies them by S^k in a way that respects their geometry. I also shoe-horned in the animated wireframes so you should be able to take a look. #189 doesn't look to be quite as bad as I thought it was but it's still pretty bad. I'll chat to Lukas about it tomorrow.

PR #328, opened 2022-01-10
Layout and mesh extraction for arbitrary dimensions

calintat:
Cubicalisation is dead, long live the new layout and mesh algorithms


doctorn:
This looks fantastic and a lot cleaner than what we had previously

doctorn:
@NickHu merging blocked because of the signing thing. Can we just turn that off?

PR #351, opened 2022-01-24
Refactor geometry extraction

### LINKED ISSUE #340: Cannot render irregular cubes
The current way we render cubes does not work for irregular cubes like the one below. It leaves gaps in the space, which is what we were seeing in #189 for the lisp.

[lips_2.hom.zip](https://github.com/homotopy-io/homotopy-rs/files/7897981/lips_2.hom.zip)
 
IMG_A2E1583F5F05-1

calintat:
* Refactored the geometry extraction and moved it into its own module `homotopy-graphics/src/geom.rs`.
* This paves the way to using the same mesh/layout/geometry pipeline in both the 2D and 3D/4D renderers.
* ~~Delayed the curve extraction until after subdivision, so curves are no longer explicitly stored in the geometry.~~
* Delayed the ordering of tris/tetras by time coordinate until buffering since this is only needed in 4D.
* Switched to the Freudenthal/6-fold triangulation which fixes #340.
* In theory, we could also make the minimal/5-fold triangulation work, but I had some issues getting that working.


calintat:
Further work (outside of this PR and needs more investigation):
* Completely bin the simplicial geometry and instead render the cubical geometry directly using Nathan's idea.
* Refactor subdivision so that it preserves direction, which I think will make the orientation stuff easier.

doctorn (approved):
This looks fantastic - check the lints and then I think we're good to go! Thanks so much Calin!!!

PR #378, opened 2022-02-03
Add debug panel

NickHu:
Would be helpful to have some tools like this. Unfortunately, it doesn't the `#[cfg(debug_assertions)]` doesn't seem to work properly - it doesn't compile in release mode because of some weirdness that I haven't been able to fix. Could you possibly help @doctorn?


doctorn:
I'll have a look now!

PR #494, opened 2022-06-07
4d occlusion

calintat:


NickHu:
So (apparently) this works mostly, but there are some glitches.

I asked a graphics programmer friend about how to do this. He said something like the following:

This hit-testing stuff assumes that each bit of geometry is drawn front-to-back (relative to the screen). The GLSL bit is probably fine, but the bit where the draw calls are being built up before it gets to the shader pipeline should incorporate some kind of sort to (at least approximately) satisfy this assumption.

Now, one of the problems is that in 4D, one wire is not going to necessarily be consistently in front or behind some other wire, and because we don't generate new meshes we can't really get around this. One way to 'fix' this is to approximate the location of the wire in 3D (by averaging over time) and sort like this; it ought to reduce the glitching, but I don't know if it will fix the problem.

The bit where we build up the draw calls is here:
https://github.com/homotopy-io/homotopy-rs/blob/e7097f265d5671417810c226c4b18edfd02b0c12/homotopy-gl/src/frame.rs#L189-L238

There's a second approach to doing this which is potentially better. The idea is to do the equivalent of sorting the draw calls, but without actually sorting, and doing it in the shader. It would be 100% accurate and also fast. Rendering should be done in two passes. In the first pass, figure out which wire is closest to the camera for each pixel, which can be done by rendering each wire (as is done currently) and using GL depth testing etc; instead of rendering to the screen, render to an intermediate framebuffer. The data to be stored is the distance between the camera and the wire (i.e. depth) and wire id at each pixel:

```glsl
struct Pixel { float depth; u32 wire_id; };
Pixel screen[w * h];
```

In the second pass, for each pixel, search in a small window if there are nearby pixels that have a different wire id that are closer to the camera; if so, then discard this pixel. This is essentially what this PR does already, but adds a prepass to make sure it works correctly.

doctorn:
> This hit-testing stuff assumes that each bit of geometry is drawn front-to-back (relative to the screen). The GLSL bit is probably fine, but the bit where the draw calls are being built up before it gets to the shader pipeline should incorporate some kind of sort to (at least approximately) satisfy this assumption.

It doesn't assume that because we use a GBuffer which sorts things using the z-buffer algorithm

> There's a second approach to doing this which is potentially better. The idea is to do the equivalent of sorting the draw calls, but without actually sorting, and doing it in the shader. It would be 100% accurate and also fast. Rendering should be done in two passes. In the first pass, figure out which wire is closest to the camera for each pixel, which can be done by rendering each wire (as is done currently) and using GL depth testing etc; instead of rendering to the screen, render to an intermediate framebuffer. The data to be stored is the distance between the camera and the wire (i.e. depth) and wire id at each pixel:

This is how it works at the moment but without ID tagging. Maybe that would be better I don't know...

NickHu:
> This is how it works at the moment but without ID tagging. Maybe that would be better I don't know...

Is it possible the artifacting we're observing is caused by wires occluding themselves, and this can be solved with ID tagging?

jamievicary:
Hi, we should not read too much into the current occlusion bug. Occlusion was working very nicely as implemented by Nathan, until Calin and I made some changes to increase the performance and the occlusion width. Our changes worked well but introduced some jankiness. This is all shader programming and there are parts of the shader program we do not even understand, and we have not even sat down once yet with Nathan to try to debug it. So please do not imagine that this is necessarily some deep problem at this stage.

doctorn:
This is by no means perfect but it's a lot better

calintat:
@doctorn This looks great, thanks Nathan! How different is your approach to what Jamie and I did. Also how easy would it be to increase the "occlusion gap" or is it hardcoded in?

doctorn:
I traded artefacts for a performance hit by testing the entire kernel rather than just the perimeter and then optimised the implementation as much as I could. To change the occlusion gap you just need to code a new mask kernel (change `MASK_SAMPLES_*`) however any bigger and this approach is too slow. There are various `HACK_*` constants you can fiddle with to improve the appearance but I found these numbers work best for medium size diagrams. On large diagrams there are still similar artefacts from certain viewing angles and I know why they occur but there is no obvious solution.

jamievicary:
Thanks Nathan this is brilliant, can we merge please.

calintat:
@doctorn I think we can merge this (unless there are more urgent changes you want to make).

Issue #525, opened 2022-06-20
Image export dialog


calintat:
When exporting a diagram to TikZ, a confirmation dialog should pop up where you can customise and preview your TikZ export:
- turn off braidings to to get a more compact TikZ export,
- black-and-white or coloured,
- generator style (#519),

For 3D diagrams, you could have the option of exporting a stack of 2D diagrams (#475). Or maybe we could export them in Manuel's style which would also work for 4D diagrams:

<img width="170" alt="Screenshot 2022-06-20 at 10 18 14" src="https://user-images.githubusercontent.com/16837293/174569692-e30bf6f6-a9c3-483c-9e05-59e9bd11f117.png">

calintat:
Now that we can also export SVGs, we need different buttons for "Export to TikZ" and "Export to SVG".

calintat:
Thanks Yulong for implementing the first version!

In our standup yesterday, we decided to promote the image export panel to a full-blown drawer panel. We therefore need to change the icon to a "Share" button, move it to the top of the sidebar, and have it always be visible. Moreover, when you open the panel and there is no diagram in the workspace, it should say something like "No diagram visible".

Finally, we need to start adding some options. For now, let's start with these two and we can think of more later:
- a toggle "Show braidings" under TikZ,
- a toggle "Use OpenGL renderer" under Manim.

jamievicary:
Why do we need this "show braidings" toggle for Tikz?

calintat:
Turning off the braidings would simplify the TikZ code and if you're working in a symmetric monoidal category, you might not care about the braidings anyway.

calintat:
And we already have the code to generate TikZ without braidings, so we might as well expose it to the user.

jamievicary:
That makes sense.

PR #556, opened 2022-06-28
SVG export, closes #196

### LINKED ISSUE #196: SVG export
there is no option to export the current cell as an image.

[wish] I would use nothing else if there were the option to export to tikzit, even tikz.

regular-citizen:
Implemented by pulling the SVG directly from the DOM, as we already render it.

This is a slightly non-trivial business as we need both the generated SVG and its stylesheets, which are handled by completely different code paths.

As documented in homotopy-web/src/model.rs, we add IDs to both the panzoom component containing the SVG and the stylesheet created by SignatureStylesheet, to the use DOM APIs to request the inner_html.

I'm not thrilled about the amount of code this adds to homotopy-web/src/model.rs (although most of it is documentations), and we definitely need a different icon for exporting TiKz and SVG.

### DIFF
--- a/homotopy-web/src/app/sidebar/buttons.rs
+++ b/homotopy-web/src/app/sidebar/buttons.rs
@@ -115,4 +115,10 @@ declare_sidebar_tools! {
         "download",
         model::Action::ExportTikz,
     }
+
+    BUTTON_SVG {
+        "Export to SVG",
+        "download",
+        model::Action::ExportSvg,
+    }
 }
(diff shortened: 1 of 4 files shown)


calintat (approved):
This is great, thanks Chiara!

I agree we could probably clean this up if we saved the rendered SVG and the signature stylesheet somewhere, but this is good enough for now.

calintat:
Once the image export dialog (#525) is done, we can have a single button in the sidebar and then in the dialog we can have separate buttons for "Export to TikZ" and "Export to SVG".

calintat:
@regular-citizen I'll leave it up to you to merge the PR, I believe you should have permission to do that.

PR #584, opened 2022-07-14
fix panzoom bug, make source/target preview smaller

H-Yulong:


calintat:
Thanks Yulong for doing this so quickly!

This fixes the panzoom and clipping issues, but now the signature drawer is narrower, can we please fix that?

Also I think the previews shouldn't be scrollable and instead we should have a max width and shrink the diagram to fit that (while maintaining the aspect ratio). The point is to be able to quickly glance at the source/target while constructing the other one.

Since the panzoom bug is more pressing, maybe we could do that in this PR (once we fix the signature drawer) and then make the previews smaller and non-scrollable in a further PR.

H-Yulong:
The signature drawer is now backed to its normal size. I'll fix other problems in a new PR.

calintat:
Thanks Yulong, we can merge this.

I'll make an issue for making the preview non-scrollable so we can keep track of that.

Issue #460, opened 2022-04-26
Render 0-cells as boxes


calintat:
As discussed in our meeting today, it would be nice to have an option to render 0-cells as boxes instead of points.

jamievicary:
You mean vertices, not 0-cells.

thud:
Thanks, that clears it up a little. I've been working on making the vertices individually restylable (ie. each vertex can be a box or a circle or some other shape). I think this will be relatively straightforward to finish but it may conflict slightly with the _global_ setting.

For now I'll keep working on individual styles. Let me know if I should refocus on the global setting instead.

calintat:
Perfect, that's even better.

doctorn:
Nice work @thud! Maybe split out a `GeneratorStyle` struct that can be shared rather than sharing the whole `GeneratorInfo`?

Let me know if you need a hand making your state changes backwards compatible with the serialiser!

thud:
> Maybe split out a `GeneratorStyle` struct that can be shared rather than sharing the whole `GeneratorInfo`?

Oh yeah, that's much better actually. I'll work that through today and see what I can get running.

calintat:
We might also need the `name` if say we decided to label the vertices, so it might be good to share the whole `GeneratorInfo` just for future convenience.

jamievicary:
We definitely need tooltips showing the generator name when we hover the
mouse cursor over the diagram.

calintat:
We already have that right? Also it's not related to this bug I think.

jamievicary:
This is a GitHub bug. I made that comment months ago by email and it only just got processed.

calintat:
Haha interesting.

PR #588, opened 2022-07-14
Vertex styles for generators

thud:
* Add generator style and related structs which allow the user to change the shape (and in future labels etc.) of different generators and their vertices. (#460)
* Add rendering for these vertices (SVG, GL, Tikz, Manim, STL)


homotopy-graphics/src/style.rs
```diff
@@ -0,0 +1,22 @@
+use homotopy_core::Generator;
+
+pub trait GeneratorStyle {
+    fn label(&self) -> Option<String>;
+    fn shape(&self) -> Option<VertexShape>;
+    // TODO(thud): migrate color-related code to here for consistency
+}
+
+pub trait GeneratorStyles<T: GeneratorStyle> {
```
calintat, line 9: This is a great way to avoid the dependency problem, well done!

Have you considered moving the type `T` inside the trait to avoid having so many generics everywhere, i.e.

```rust
pub trait GeneratorStyles {
    type T: GeneratorStyle;

    fn generator_style(&self, g: Generator) -> Option<&Self::T>;
}
```

Also (and this is more of a nitpick) could we find a different name? How about `Signature` or `SignatureStylesheet`?

homotopy-graphics/src/geom.rs
```diff
@@ -448,9 +500,21 @@ impl SimplicialGeometry {
         }
     }
 
-    pub fn inflate_3d(&mut self, samples: u8) {
+    pub fn inflate_3d<S, T>(&mut self, samples: u8, generator_styles: Option<&S>)
```
calintat, line 503: I don't think `generator_styles` needs to be `Option`.

homotopy-graphics/src/manim.rs
```diff
@@ -13,16 +13,52 @@ use itertools::Itertools;
...
-use crate::svg::render::GraphicElement;
+use crate::{
+    style::{GeneratorStyle, GeneratorStyles, VertexShape},
+    svg::render::GraphicElement,
+};
 
 const OCCLUSION_DELTA: f32 = 0.2;
 const INDENT: &str = "    ";
 
+trait ManimRenderVertex {
```
calintat, line 24: I would maybe replace this single-use trait with a top-level function:

```rust
fn render_vertex(generator_style: GeneratorStyle, color: &str) -> String
```

homotopy-graphics/src/manim.rs
```diff
@@ -13,16 +13,52 @@ use itertools::Itertools;
...
+}
+
 pub fn color(generator: Generator) -> String {
     format!("generator_{}_{}", generator.id, generator.dimension)
 }
 
-pub fn render(diagram: &Diagram, stylesheet: &str) -> Result<String, DimensionError> {
+pub fn render<S, T>(
+    diagram: &Diagram,
+    generator_styles: Option<&S>,
```
calintat, line 55: Again `generator_styles` doesn't need to be an `Option`.

homotopy-graphics/src/style.rs
```diff
@@ -0,0 +1,22 @@
+use homotopy_core::Generator;
+
+pub trait GeneratorStyle {
+    fn label(&self) -> Option<String>;
+    fn shape(&self) -> Option<VertexShape>;
+    // TODO(thud): migrate color-related code to here for consistency
```
calintat, line 6: Yes, please do that!

homotopy-graphics/src/style.rs
calintat, line 9: This is a great way to avoid the dependency problem, well done!

Have you considered moving the type `T` inside the trait to avoid having so many generics everywhere, i.e.

```rust
pub trait GeneratorStyles {
    type T: GeneratorStyle;

    fn generator_style(&self, g: Generator) -> Option<&Self::T>;
}
```

Also could we find a different name? How about `Signature` or `SignatureStylesheet`?

homotopy-graphics/src/geom.rs
calintat, line 503: I think it would be better if `generator_styles` wasn't an `Option`.

homotopy-graphics/src/stl.rs
```diff
@@ -1,17 +1,27 @@
...
 
-pub fn render(diagram: &Diagram) -> Result<String, DimensionError> {
+use crate::{
+    geom::{CubicalGeometry, SimplicialGeometry},
+    style::{GeneratorStyle, GeneratorStyles},
+};
+
+pub fn render<S, T>(
+    diagram: &Diagram,
+    generator_styles: Option<&S>,
```
calintat, line 12: Again this doesn't need to be an `Option`.

homotopy-graphics/src/tikz.rs
```diff
@@ -13,15 +13,59 @@ use itertools::Itertools;
...
 
-use crate::svg::render::GraphicElement;
+use crate::{
+    style::{GeneratorStyle, GeneratorStyles, VertexShape},
+    svg::render::GraphicElement,
+};
 
 const OCCLUSION_DELTA: f32 = 0.2;
 
+trait TikzRenderVertex {
```
calintat, line 23: Again we don't need this trait.

homotopy-graphics/src/tikz.rs
```diff
@@ -13,15 +13,59 @@ use itertools::Itertools;
...
+
 pub fn color(generator: Generator) -> String {
     format!("generator-{}-{}", generator.id, generator.dimension)
 }
 
-pub fn render(diagram: &Diagram, stylesheet: &str) -> Result<String, DimensionError> {
+pub fn render<T, S>(
+    diagram: &Diagram,
+    stylesheet: &str,
+    generator_styles: Option<&T>,
```
calintat, line 63: Again this doesn't need to be an `Option`.

homotopy-graphics/src/tikz.rs
```diff
@@ -97,10 +141,18 @@ pub fn render(diagram: &Diagram, stylesheet: &str) -> Result<String, DimensionEr
     }
 
     // Points
+    // TODO(thud): this `default_shape` should not be hardcoded here
```
calintat, line 144: You can get rid of the default shape if `generator_styles` isn't an `Option`.

homotopy-web/src/app.rs
```diff
@@ -89,6 +89,19 @@ impl Component for App {
                     }
                 }
 
+                // use model::proof::{Action::EditSignature, SignatureEdit::Edit};
```
calintat, line 92: Please remove these before merging.

homotopy-web/src/model/proof.rs
```diff
@@ -167,7 +168,9 @@ impl Action {
     pub fn relevant(&self) -> bool {
         !matches!(
             self,
-            Action::HighlightSlice(_) | Action::HighlightAttachment(_)
+            Action::HighlightSlice(_)
+                | Action::HighlightAttachment(_)
+                | Action::EditSignature(SignatureEdit::Edit(_, _))
```
calintat, line 173: This is not quite right. This means pressing `undo` will undo both the signature edit and whatever action came before that. I think it's better to leave it as it was.

calintat (commented):
This looks brilliant, thanks so much!

I left some comments, most are them are minor.

homotopy-web/src/model/proof.rs
thud: Yeah that makes sense, easy fix.

homotopy-graphics/src/style.rs
thud: In [c9f325d](https://github.com/homotopy-io/homotopy-rs/pull/588/commits/c9f325ddc6648f59b6bda531f44e4f60fe206ee2), I have renamed `GeneratorStyles` to `SignatureStyleData`.

I think `SignatureStylesheet` might get confusing quickly in the current codebase and `Signature` is perhaps too vague here.

If we merged `SignatureStylesheet` code into `SignatureStyleData`, then we could think about another rename.

homotopy-graphics/src/style.rs
thud: I think it would be possible to merge the `SignatureStylesheet` functionality into this trait which would make the codebase more consistent and clean.

calintat: Yes, that would be great. Let's do that in a future PR.

homotopy-graphics/src/tikz.rs
calintat: You can just `unwrap` here. If the generator isn't in the signature, something has gone wrong.

calintat (approved):

homotopy-graphics/src/tikz.rs
thud: Great, that's sorted now. There was something similar in the manim rendering engine which I have now also fixed.

Issue #538, opened 2022-06-20
Merge wires in 2D


calintat:
This should be fairly easy and would make the SVG/TikZ code smaller. Also see how we merge curves in 3D/4D.

Issue #533, opened 2022-06-20
TikZ exported code unclean


NickHu:
Currently, the TikZ export will name nodes things like `generator-0-0`, when a better choice would be e.g. the name of the generator.
It also exhibits floating point precision ugliness by giving coordinate values like `5.000000000000000027` which should be rounded to a 'nice' number of significant figures.

calintat:
Weird I thought I had fixed the rounding problem.

calintat:
But I agree that the code can be cleaner. The only obstacle to using the generator name as the name of the colour is that you could have multiple generators with the same name.

NickHu:
> Weird I thought I had fixed the rounding problem.

Perhaps this is true; I haven't checked recently.

> But I agree that the code can be cleaner. The only obstacle to using the generator name as the name of the colour is that you could have multiple generators with the same name.

Easy to fix! Probably `cup-0` and `cup-1` are still better than 'generator-X-Y` in my opinion.

jamievicary:
Chiara good for you to look at?

regular-citizen:
Merged PR to reduce the amount of rendering primitives issued.
I will make another PR to actually clean up the tikz code.

PR #591, opened 2022-07-15
Path simplification and wire merging

### LINKED ISSUE #537: Optimise surface merging in 2D
Currently surfaces contain many collinear points which makes the SVG and TikZ code more complicated that it needs to be.

Note that this is a geometric optimisation, it cannot be done at the logical level as it depends on the layout.

regular-citizen:
This is PR is Part 1/2 of work to clean up the output of (human readable) exporters, e.g. issue #533.

It add a path_util.rs, which contains two key functions:
- The first one is the path simplifier: it takes lyon paths and removes unnecessary drawing primitives. This is very easy to call in a stand-alone fashion from various renderers.
- The other is the graphic simplifier. This this consumes the entire GraphicElement vector to be rendered, and does some more drastic simplifications by merging wires, while respecting the masking. Note that this function is mostly meant for the tikz/manim exports, as it discards the SVG masking!

This closes #537 and #538, and partly addresses #533 (Eckmann-Hilton .tikz export has been reduced by circa 60%).

### DIFF
--- a/homotopy-graphics/src/path_util.rs
+++ b/homotopy-graphics/src/path_util.rs
@@ -10,30 +10,28 @@ pub fn simplify_graphic<const N: usize>(graphic: &[GraphicElement<N>]) -> Vec<Gr
     let mut new_graphic = Vec::with_capacity(graphic.len());
     let mut point_elements = Vec::new();
 
-    // depth -> (mask, last_point, Vec<(path, gen)>)
+    // (depth, gen) -> Vec<path>
     let mut grouped_wires =
-        FastHashMap::<usize, (Vec<Path>, Vec<(lyon_path::path::Builder, Generator)>)>::default();
+        FastHashMap::<(usize, Generator), Vec<lyon_path::path::Builder>>::default();
 
     for element in graphic {
         match element {
             GraphicElement::Surface(g, path) => {
                 new_graphic.push(GraphicElement::Surface(*g, simplify_path(path)));
             }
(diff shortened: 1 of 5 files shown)


homotopy-graphics/src/path_util.rs
```diff
@@ -0,0 +1,368 @@
...
+                            extremes.0.approx_eq(to),
+                            extremes.1.approx_eq(to),
+                            extremes.1.approx_eq(from),
+                            extremes.0.approx_eq(from),
+                        ) {
+                            (true, _, _, _) => {
+                                builder.extend_from_paths(&[path.as_slice()]);
+                                *to = extremes.1;
+                            }
+                            (_, true, _, _) => {
```
calintat, line 54: Could we please drop this case? It feels wrong to reverse paths, and it will definitely cause problems when we add arrows to wires.

homotopy-graphics/src/path_util.rs
```diff
@@ -0,0 +1,368 @@
...
+                            (_, _, true, _) => {
+                                let mut new_builder = Path::builder();
+                                new_builder.extend_from_paths(&[
+                                    path.as_slice(),
+                                    builder.clone().build().as_slice(),
+                                ]);
+                                *builder = new_builder;
+                                *from = extremes.0;
+                            }
+                            (_, _, _, true) => {
```
calintat, line 71: This one too I think.

calintat (approved):
This looks great Chiara, thanks a lot! I just have one comment about reversing paths.

PR #593, opened 2022-07-18
Tikz cleanup (#533)

regular-citizen:
This PR is aimed at addressing #533.
- It deprecates the current offsetting strategy, in favour of doing the masking entirely in tikz.
- It incorporates a series of changes aimed at improving the readability of the output.
- It documents better the exporting procedure.

Feedback on the output is greatly appreciated.


calintat:
Also since each path will only appear in one layer, maybe we can inline them and get rid of the roman numbers. I don't know.

calintat:
This looks brilliant Chiara, thanks so much!

Btw can we rename `layered` to `wire` and `clippedlayer` to `layer`? I think those names would be more intuitive.

calintat (approved):

NickHu:
Rebased

PR #595, opened 2022-07-19
Make generator style changes immediate

thud:


thud:
Note that CI linting failures seem to be due to a bug with the nix package and a newer version of rustc nightly which requires a new dependency.

Issue #589, opened 2022-07-15
Making generators invertible and framed/oriented


calintat:
## Invertible

Each generator should have a checkbox labelled "Invertible" which will be off by default. This toggles the `invertible` field in that generator's `GeneratorInfo` which determines whether you are allowed to attach its inverse or not.

This is basically already implemented, we just need to do the UI.

## Framed / oriented

This is harder as you can only move in one direction: framed > oriented.

Here is what I am proposing:
- when you add a new generator, it is framed by default.
- you can make it oriented which will trigger a procedure that removes all the framing information from every occurrence of that generator in the signature or workspace (basically check if the label belongs to that generator, and if so set it to null).
- once a generator is oriented, you are not allowed to make it framed again.

In terms of UI, I think we should have a drop-down menu that lets you choose between framed and oriented (and in the future unoriented). Once you have selected oriented, framed should be greyed out to tell the user they cannot go back.

calintat:
In the future, we will want more checkboxes for things like signature previews, so this will be a great foundation for that.

thud:
Some todos for future:
- [ ] Toggleable previews option...
- [ ] Framed / Oriented option...
- [ ] Show label option...

I'll be working on `SignatureStyleData` related code in the meantime.

calintat:
Thanks Jasper, I will do the framed/oriented toggle and then we can close this.

PR #602, opened 2022-07-20
Better style picker

thud:
This changes the functionality of the generator style-picking UI to allow for preferences such as invertibility (#589).
It also eliminates the intermediate 'edit' state from this UI for simplicity.


calintat:
Amazing!

Just a few comments:
- can we move invertibility to the top,
- can we make the whole region around the text/checkbox clickable
- is it possible to disable the checkbox on certain generators (e.g. 0-generators)?

thud:
No problem, shouldn't take long...

calintat:
Sorry I broke your branch :(

thud:
I've pushed a fix for these features.

This removes the invertibility option entirely from 0d generators. If preferred, I can change to keep displaying but disable input.

calintat:
Brilliant, thanks Jasper!

I think it's better to remove the checkbox, rather than disabling it.

thud:
Great

Issue #586, opened 2022-07-14
Make boundary previews non-scrollable


calintat:
I think the source/target previews shouldn't be scrollable and instead we should have a max width and shrink the diagram to fit that (while maintaining the aspect ratio).

calintat:
This is basically implemented, blocked by #609.

calintat:
Fixed by https://github.com/homotopy-io/homotopy-rs/commit/1b98300eac44c21c69fa5a255a34c2609fd55d9d.

PR #604, opened 2022-07-21
Improve boundary preview

H-Yulong:
Now you can:
- Click on the "Source/Target" text to switch between source/target
- Click on the diagram to recover it (if the workspace is empty)
- Remove the source/target


calintat:
Thanks Yulong, this looks nice!

calintat:
This looks great, thanks Yulong! I think we can merge this now and will hopefully fix #609 soon.

H-Yulong:
Great! I think it might be better to close #586 after #609 is fixed.

homotopy-web/src/app/boundary.rs
```diff
@@ -78,6 +100,7 @@ impl BoundaryPreview {
                     diagram={ctx.props().boundary.diagram.clone()}
                     id="boundary__preview"
                     signature={ctx.props().signature.clone()}
+                    dimensions={Some((160.0, 160.0))}
```
calintat, line 103: Can you please rename this to `max_dimensions`? Or maybe split it into `max_width` and `max_height`?

H-Yulong: Sure, just committed.

calintat: Thanks!

Issue #601, opened 2022-07-20
Signature stylesheet refactor


calintat:
The code related to colours and generating the signature stylesheet should be migrated to `homotopy_graphics` and be part of `GeneratorStyle` and/or `SignatureStyleData`.

Also for the TikZ export, we want to use the generator name as the colour name instead of the generic `generator_id_dim` (supersedes #533).

Issue #550, opened 2022-06-24
Orientation information in the renderer


calintat:
Once the orientation labelling (#243) is done, we need to display this information in the renderered diagram.

Here are some ideas:
- different colours (how will this interact with the different shades for points, wires, regions)
- asymmetric boxes for the points (related to #519)
- arrows on the wires (#553)
- dashed/dotted wires?
- flip normals in GL

jamievicary:
Hi -- can this please be prioritised, now we are using the alpha tool a lot it would be really great to get this rendering information in properly.

jamievicary:
There is a problem with this idea unfortunately. Basically we cannot change the saturation, since a HSL value with positive saturation, and non-extremal lightness, will always have some "colour" to it, i.e. it will not be on the black-white spectrum. This means that, if we varied S/L as we were planning, a white or black generator would sometimes be shown with a nontrivial hue.

I think the best alternative is to just use 9 different lightness values. For example:

    L=15%, vertex, orientation -1
    L=23%, vertex, orientation 0
    L=31%, vertex, orientation +1
    L=39%, wire, orientation -1
    L=47%, wire, orientation 0
    L=55%, wire, orientation +1
    L=63%, surface, orientation -1
    L=71%, surface, orientation 0
    L=79%, surface, orientation +1

We may want to cycle these choices if the initially-chosen colour was particularly light or dark.

calintat:
Is it really worth differentiating surfaces? For points and wires it makes sense because you can contract those, but for surfaces it seems superfluous. That would at least reduce the number of lightness values.

jamievicary:
Yes I think it's important because you can have homotopies-on-homotopies-on-homotopies of the same generator. So you can have an entire diagram of points, wires and surfaces all with the same generator label (and potentially all with the same orientation).

thud:
Alright I've set the lightness according to this in #717.
> L=15%, vertex, orientation -1
> L=23%, vertex, orientation 0
> L=31%, vertex, orientation +1
> L=39%, wire, orientation -1
> L=47%, wire, orientation 0
> L=55%, wire, orientation +1
> L=63%, surface, orientation -1
> L=71%, surface, orientation 0
> L=79%, surface, orientation +1

I think this is going to require a little more tweaking given that some of the
colours seem far too dark etc.
I'm going to play around with the colours and see if I can get something which
is clear but sufficiently different between codimensions/orientations.

thud:
[...]

PR #608, opened 2022-07-22
Integrate signature styling into homotopy-graphics

thud:
This addresses (#601)


calintat:
Thanks Jasper, can you please rebase your commits on top of master?

thud:
Thanks

homotopy-graphics/src/style.rs
```diff
@@ -1,15 +1,26 @@
...
 
+use crate::tikz;
+
 pub trait GeneratorStyle {
+    type Color: RenderableColor;
+
     fn label(&self) -> Option<String>;
     fn shape(&self) -> Option<VertexShape>;
-    // TODO(thud): migrate color-related code to here for consistency
+    fn color_point(&self) -> Self::Color;
```
calintat, line 12: I reckon this should contain a single `color` function.

homotopy-web/src/model/proof/generators.rs
```diff
@@ -38,11 +44,38 @@ impl GeneratorStyle for GeneratorInfo {
     fn shape(&self) -> Option<style::VertexShape> {
         Some(self.shape.clone().into())
     }
+
+    fn color_point(&self) -> Self::Color {
+        self.color.clone()
+    }
+
+    fn color_wire(&self) -> Self::Color {
```
calintat, line 52: This could move into `homotopy_graphics` and it can be an impl function on `GeneratorStyle.

homotopy-web/src/model/proof/generators.rs
```diff
@@ -38,11 +44,38 @@ impl GeneratorStyle for GeneratorInfo {
...
+        Color((self.color.into_linear().lighten(0.05)).into())
+    }
+
+    fn color_surface(&self) -> Self::Color {
```
calintat, line 56: Same as above.

homotopy-graphics/src/style.rs
```diff
@@ -1,15 +1,26 @@
...
 }
 
 pub trait SignatureStyleData {
-    type T: GeneratorStyle;
+    type Style: GeneratorStyle;
 
-    fn generator_style(&self, g: Generator) -> Option<&Self::T>;
+    fn generator_style(&self, g: Generator) -> Option<&Self::Style>;
+
+    // It would be nice if the following could be an iterator but the generics get complex fast
```
calintat, line 22: I think this is unavoidable sadly :(

homotopy-graphics/src/style.rs
```diff
@@ -22,3 +33,78 @@ impl Default for VertexShape {
...
+    fn css(&self) -> String;
+    fn tikz(&self) -> String;
+}
+
+pub trait CssStylesheet {
+    fn css_stylesheet(&self, prefix: &str) -> String;
+    fn css_class(prefix: &str, generator: Generator, suffix: &str) -> String;
+}
+
+pub trait TikzStylesheet {
```
calintat, line 47: Could you please move this into the tikz module?

Also it doesn't have to be a trait, it could just be a top-level function, but that's up to you.

homotopy-graphics/src/style.rs
```diff
@@ -22,3 +33,78 @@ impl Default for VertexShape {
         Self::Circle
     }
 }
+
+pub trait RenderableColor {
+    fn css(&self) -> String;
+    fn tikz(&self) -> String;
+}
+
+pub trait CssStylesheet {
```
calintat, line 42: Could you please move this into the svg module? It can be either in `svg.rs` or another file inside the `svg` directory.

homotopy-web/src/model/proof/generators.rs
```diff
@@ -38,11 +44,38 @@ impl GeneratorStyle for GeneratorInfo {
...
+        Color((self.color.into_linear().lighten(0.1)).into())
+    }
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct Color(pub Srgb<u8>);
 
+impl RenderableColor for Color {
```
calintat, line 64: Related to the previous comments, these should really live in the svg/tikz modules in `homotopy_graphics`. And I don't think we need the trait.

calintat (commented):
This looks really good Jasper.

I left a few comments, mostly about the code structure. I think it's important we keep the rendering pipelines contained in their own submodules in `homotopy_graphics`.

homotopy-graphics/src/style.rs
calintat: I think `svg.rs` is a good place, because I am thinking of moving the SVG generation in there as well.

homotopy-web/src/model/proof/generators.rs
thud: Yup, that sounds wise. I was trying to keep `palette` from becoming a dependency of `homotopy_graphics` but in heinsight I can probably keep all `palette` related code in `homotopy_graphics` instead.

So I'll try removing `palette` from `homotopy-web` dependencies and adding it to `homotopy_graphics` dependencies.

thud: This should clear up the comments above

calintat: Yes, the `palette` dependency can move to `homotopy_graphics`.

homotopy-graphics/src/style.rs
thud: Yeah that makes sense. The `css_stylesheet` method is also used by `SignatureStylesheet` in `app.rs`.

I think, if I do this right, we can remove that struct altogether.

homotopy-graphics/src/style.rs
thud: Great, will do...

homotopy-graphics/src/style.rs
thud: As mentioned in one of the other comments, I think this can be fixed by moving the `palette` dependency from `homtopy_web` into `homotopy_graphics`.

homotopy-graphics/src/style.rs
calintat: That would be great!

thud:
I think that very soon we should be able to remove all top-level calls (in `homotopy_web`) to stylesheet generation. Instead, it can be handled entirely by `homotopy_graphics`. This would clean up `app.rs` and `model.rs` a little.

This would involve generating stylesheets in the `render` function in each of `tikz`, `svg`, `manim` etc.

thud:
Great, yeah that's a good point. I think it only really applies to the `svg` rendering really since the other stylesheets are generated only on export which are infrequent.

calintat:
I think it might be better to keep the rendering and stylesheet generation as separate functions, at least for SVG. But they should both be in `homotopy_graphics`.

homotopy-graphics/src/style.rs
```diff
@@ -1,20 +1,66 @@
...
 
 pub trait SignatureStyleData {
-    type T: GeneratorStyle;
+    type Style: GeneratorStyle;
+
+    fn generator_style(&self, g: Generator) -> Option<&Self::Style>;
 
-    fn generator_style(&self, g: Generator) -> Option<&Self::T>;
+    // It would be nice if the following could be an iterator but the generics get complex fast
+    fn generators(&self) -> Vec<Generator>;
```
calintat, line 18: If you look at where we use this, it's always the same pattern:

```rust
for g in signature_styles.generators() {
    let style = signature_styles.generator_style(g);
    ...
}
```

But `generator_style` expands to `generator_info` which works by iterating over the generators again until it finds the right one. So the whole thing is quadratic time.

We can make this better if we make this function return a list of `Style` rather than `Generator`.

calintat: And probably change the name to `generator_styles` or something.

homotopy-graphics/src/svg.rs
```diff
@@ -1,2 +1,51 @@
...
+}
+
+pub fn class(prefix: &str, generator: Generator, suffix: &str) -> String {
+    format!(
+        "{}__{}-{}--{}",
+        prefix, generator.id, generator.dimension, suffix
+    )
+}
+
+fn hex(color: &Color) -> String {
```
calintat, line 49: Could we move this into `Color`? Since it seems we use it in quite a few places.

homotopy-graphics/src/svg.rs
```diff
@@ -1,2 +1,51 @@
+use std::fmt::Write;
+
+use homotopy_core::Generator;
+
+use crate::style::{Color, GeneratorStyle, SignatureStyleData};
+
 pub mod render;
 pub mod shape;
+
+pub fn stylesheet(styles: &impl SignatureStyleData, prefix: &str) -> String {
```
calintat, line 10: Let's get rid of the prefix, it's always "generator".

homotopy-graphics/src/svg.rs
```diff
@@ -1,2 +1,51 @@
...
+            name = class(prefix, generator, "point"),
+            color = hex(&style.color())
+        )
+        .unwrap();
+    }
+
+    stylesheet
+}
+
+pub fn class(prefix: &str, generator: Generator, suffix: &str) -> String {
```
calintat, line 42: Looking at this, it seems we have the same function in manim, tikz, and svg (with the difference that in svg it takes a suffix as well). Maybe we could combine them and move them into `style.rs`.

calintat: Also I wanted to ask your thoughts about this idea I have had.

Instead of having three CSS classes for every generator
```css
.generator_0_0--point
.generator_0_0--wire
.generator_0_0--surface
```
would it be possible to do this by mixing CSS classes, so something like
```css
.generator_0_0
.generator_0_0 .wire
.generator_0_0 .surface
```

Is it possible to write classes `.wire` and `.surface` which lighten the stroke/fill colour?

homotopy-web/src/app/signature/item.rs
```diff
@@ -146,7 +146,10 @@ struct CustomRecolorButtonProps {
 
 #[function_component(CustomRecolorButton)]
 fn custom_recolor_button(props: &CustomRecolorButtonProps) -> Html {
-    let hex = format!("#{:X}", props.value.0);
+    let hex = {
```
calintat, line 149: This could also use the `hex` function on `Color` that I suggested in a comment above.

homotopy-web/src/app/signature_stylesheet.rs
```diff
@@ -1,16 +1,15 @@
...
+
+// It would be nice if we could remove the `SignatureStylesheet` entirely but we still need to
+// maintain a handle to the html node for mounting/unmounting. This also means we can avoid
+// regenerating the stylesheet unless the signature has changed [`update`].
 
-// TODO: Check if there is a performance problem with this. If so, then use the
+// TODO: Check if there is a performance problem with the current approach.. If so, then use the
 // stylesheet API to change the colors more granularly.
 
 pub struct SignatureStylesheet {
```
calintat, line 15: We probably can't get rid of this yet, but this is much better.

homotopy-web/src/model.rs
```diff
@@ -153,18 +140,14 @@ impl State {
...
-                    .expect("no document")
-                    .get_element_by_id("signature__stylesheet")
-                    .expect("no stylesheet in document");
-                let stylesheet = style_element.outer_html();
+                // Failure to do so gives a fully-black SVG. We can generate the stylesheet in the
+                // same way the `SignatureStylesheet` struct does.
+                // We also strip the styles of whitespace since it is unneeded.
+                let stylesheet = {
+                    let mut inner_stylesheet = svg::stylesheet(&signature, "generator");
+                    inner_stylesheet.retain(|c| !c.is_whitespace());
```
calintat, line 148: The whitespaces might be nice for readability. It might be worth keeping them for the export (since people might want to look at the svg code if they downloaded it).

calintat (approved):
This looks great, thanks Jasper!

homotopy-graphics/src/style.rs
calintat: Sorry it should return pairs `(Generator, Style)` as we need the generator as well.

homotopy-graphics/src/lib.rs
```diff
@@ -21,7 +21,7 @@
     clippy::unneeded_field_pattern,
     clippy::unreachable,
     clippy::use_debug,
-    clippy::use_self,
+    // clippy::use_self, // This gives false positives with serde derives
```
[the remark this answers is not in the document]
thud: Great, thanks

homotopy-graphics/src/style.rs
thud: Yup, good point

homotopy-web/src/model.rs
thud: Yeah, I agree though the rest of the svg code is all a single line. Perhaps we could also make _that_ multiline as well for the same reason.

homotopy-graphics/src/svg.rs
thud: For now, I have switched to using `to_string` calls instead. This results in slightly less code but if we ever change the way we `fmt::Display` colors, it could break some of the styles.

homotopy-graphics/src/svg.rs
thud: Good idea, I'll have a look at this.

homotopy-graphics/src/svg.rs
calintat: That's fine. Maybe write a comment so we don't forget.

homotopy-graphics/src/svg.rs
calintat: Thanks! That can be a separate PR.

Also related to #550, which is one of the reasons I want to do this so we don't have 9 css classes for every generator.

homotopy-web/src/model.rs
calintat: Yes good point, I will fix that as part of #609.

homotopy-graphics/src/svg.rs
thud: Great.

homotopy-graphics/src/style.rs
```diff
@@ -49,6 +49,8 @@ impl FromStr for Color {
 }
 
 // Reverse of `FromStr`
+// NOTE: If we decide to change how we `fmt::Display`, it could break some styles as it is used
```
doctorn, line 52: Maybe make a specific method for manim export colour formatting to avoid this?

thud: Yeah I think that's probably safer.
Orignally, it was implemented like that but we found that we ended up with a bunch of repeated definitions for the `hex` functions for each renderer.
In a previous comment @calintat requested this be reduced into one canonical `impl` method.

I'm honestly very happy to create those functions again for clarity. I think the easiest fix would probably be to create an actual `impl` method `hex` on top of Color and that way there is no dispute over which format should be returned.

thud: This is what I have implemented for now. As far as I can tell, this is otherwise ready to merge so I'll wait for confirmation or further comment before squash and merge.

calintat: This is better, thanks!

PR #632, opened 2022-07-28
Substantial improvements to signature UI

thud:
This should allow the user to (optionally) preview each generator in the signature panel.

It should allow the user to toggle between generator previews and source/target previews.

### DIFF
--- a/homotopy-graphics/src/style.rs
+++ b/homotopy-graphics/src/style.rs
@@ -36,6 +36,13 @@ impl Color {
         Self(palette::Lighten::lighten(self.0.into_linear(), amount).into())
     }
 
+    pub fn is_light(&self) -> bool {
+        palette::RelativeContrast::get_contrast_ratio(
+            palette::Srgb::new(1., 1., 1.),
+            self.0.into_format::<f32>(),
+        ) < 1.5
+    }
+
     pub fn into_components<T>(self) -> (T, T, T)
     where
         T: palette::stimulus::FromStimulus<u8>,
--- a/homotopy-web/src/app.rs
+++ b/homotopy-web/src/app.rs
@@ -1,5 +1,5 @@
 use boundary::BoundaryPreview;
-use settings::AppSettings;
+use settings::{AppSettings, AppSettingsKey};
 use sidebar::Sidebar;
 use signature_stylesheet::SignatureStylesheet;
 use wasm_bindgen::{closure::Closure, JsCast};
--- a/homotopy-web/src/app/settings.rs
+++ b/homotopy-web/src/app/settings.rs
@@ -14,6 +14,7 @@ declare_settings! {
         subdivision_depth: u32 = 2,
         geometry_samples: u32 = 10,
 
+        show_previews: bool = true,
         orthographic_3d: bool = false,
         specularity: u32 = 25,
         shininess: u32 = 64,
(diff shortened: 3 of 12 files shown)


calintat:
Thanks Jasper! This is looking great, we just need to tweak the UI now.
- We need a checkbox "Single image" for toggling between a single image and source/target.
- The toggle should be disabled in 0D, default to true in 1D and 2D, and default to false in 3D and above.
- Can we move the dimension between the settings icon and name so everything except the preview is left aligned.
- Also can you try to make the settings icon a light/dark grey depending on the generator colour.
- We definitely need to increase the size of the previews as they are too small right now.
- Another thing, it would be nice to have a toggle to disable the previews altogether so you can have a more compact view. I don't know if this should be a global toggle for the signature or per generator. I am favour of the former, but anyway this is not a priority right now.

thud:
0449227 addresses a few of these, though more is still needed to be done.

thud:
Written update (post standup with JV): Signature previews are good to stay small (some indication of invertible/framed/oriented would be good).

doctorn:
I just tested this out -- amazing stuff!! I did notice that it looks like you've added a settings cog to the folders as well as the generators? Just wanted to check that was intentional?

thud:
Yeah thanks, the folders could do with some rework. The colours are also pretty horrible at the moment so hopefully it can look a lot nicer. I do think we need the cog for renaming purposes (disables draggable).

thud:
Note that preview caching is currently too aggressive.

thud:
As of now, caching is overly-careful rather than overly-aggressive ensuring correctness.

calintat:
This is looking really good Jasper. Just a few comments and then we should merge this:
- We should default to "single image" in 1D/2D and "source/target" in 3D and above. We could even disable "source/target" altogether in 1D/2D as it doesn't really tell you much and it's a bit ugly.
- The unicode arrow symbol doesn't work on any of my machines :( Tbh I'm not so keen on the symbol, I would suggest removing it.
- I don't like the idea of using boxes when running out of colours. That should be something the user toggles on purpose.
- Can we move the drawer resizing into a separate PR? It looks like it might need some more polishing and we shouldn't wait on that to merge this PR.

thud:
Great, good to hear it, I'll look to get this merged imminently.

> We should default to "single image" in 1D/2D and "source/target" in 3D and above. We could even disable "source/target" altogether in 1D/2D as it doesn't really tell you much and it's a bit ugly.
- [ ] Great sounds good, I'll commit that now.

> The unicode arrow symbol doesn't work on any of my machines :( Tbh I'm not so keen on the symbol, I would suggest removing it.
- [ ] Good

> I don't like the idea of using boxes when running out of colours. That should be something the user toggles on purpose.
- [ ] Easy fix, apologies

> Can we move the drawer resizing into a separate PR? It looks like it might need some more polishing and we shouldn't wait on that to merge this PR.
- [ ] Yeah fair enough, it's still definitely experimental.

thud:
Great, good to hear it, I'll look to get this merged imminently.

> We should default to "single image" in 1D/2D and "source/target" in 3D and above. We could even disable "source/target" altogether in 1D/2D as it doesn't really tell you much and it's a bit ugly.
- [ ] Great sounds good, I'll commit that now.

> The unicode arrow symbol doesn't work on any of my machines :( Tbh I'm not so keen on the symbol, I would suggest removing it.
- [ ] Good

> I don't like the idea of using boxes when running out of colours. That should be something the user toggles on purpose.
- [ ] Easy fix, apologies

> Can we move the drawer resizing into a separate PR? It looks like it might need some more polishing and we shouldn't wait on that to merge this PR.
- [ ] Yeah fair enough, it's still definitely experimental.

calintat:
Thanks Jasper. Feel free to merge this if you want.

Issue #509, opened 2022-06-17
Compile a fast LP solver to Wasm


calintat:
The best LP solvers such as CBC or Highs are written in C/C++. So if we want to use them in our project (#508), we need to compile them in Wasm using `emscripten`. Nick made some progress on this.

jamievicary:
This is great. Has it improved the performance of the layout algorithm?

regular-citizen:
To be determined, there are some issues to address with layout tests that currently depend on minilp to address before we can move to good_lp+highs.

jamievicary:
Thanks Chiara. So is this new solver currently being used in the production
layout code?

regular-citizen:
Not yet, but the idea is to either use this solver only for production or to just bin the tests altogether.
I think @calintat is the best position to figure out what to do on this front, but once that is sorted moving to highs should be relatively straightforward.

PR #635, opened 2022-07-28
HiGHS integration (#509)

regular-citizen:
X


regular-citizen:
Happy to merge at this point.

I think there is a reasonable way to be able to keep layout in core: if all minilp code is switched to good_lp, we could activate the highs feature during build for the web and the minilp feature during test/benchmark builds. So long as testing isn't asking for crazy layouts, this should be fine, and it may well be worth it to spare the hassle of moving layout into graphics.
Do note that this is as easy as writing "test features" in the toml, as cargo does not support activating features only for test/benchmarks, so we would need to pass a --features flag when we test.

Another option to keep layout in core would be to switch to testing in headless browser, which is supported by wasm-bindgen.

calintat:
> Happy to merge at this point.
> 
> I think there is a reasonable way to be able to keep layout in core: if all minilp code is switched to good_lp, we could activate the highs feature during build for the web and the minilp feature during test/benchmark builds. So long as testing isn't asking for crazy layouts, this should be fine, and it may well be worth it to spare the hassle of moving layout into graphics. Do note that this is as easy as writing "test features" in the toml, as cargo does not support activating features only for test/benchmarks, so we would need to pass a --features flag when we test.
> 
> Another option to keep layout in core would be to switch to testing in headless browser, which is supported by wasm-bindgen.

The first option sounds great! The tests are fairly basic and are already running fine with minilp.

regular-citizen:
> The first option sounds great! The tests are fairly basic and are already running fine with minilp.

Yeah it should be fairly reasonable to do. Things to watch out for are CI testing builds and git hooks.

calintat:
Tbh I am also happy to bin the layout tests if it comes to that.

regular-citizen:
> Tbh I am also happy to bin the layout tests if it comes to that.

Up to you. Remember to remove the dependency on good_lp from homotopy-web when you get around doing that.

regular-citizen:
> Tbh I am also happy to bin the layout tests if it comes to that.

Thinking about it, you could feature gate them and keep them around. If the feature "minilp" is present include them and not otherwise. This doesn't break default test builds but leaves you the option of running them if you want.

Issue #645, opened 2022-08-01
Changing generator name issues an history event for each character


regular-citizen:
Changing the name of a generator spams the history with all the character by character changes.
It should issue a single history event with the change of generator name.

Steps to reproduce:
- Add a generator
- Rename generator to a long name
- Undo repeatedly

calintat:
Has this changed recently? Because I don't remember this being the case.

regular-citizen:
It is definitely the case right now. If you're positive that this was introduced recently, we can figure out the commit that caused this by bisecting.

calintat:
Yeah that's good idea.

regular-citizen:
Regression introduced at 27c0f7e486fff9323b4aa1d693293b0420fd1364 by @thud.

thud:
Yep, this is an artifact of instant style updates. I missed this originally but should be a relatively easy fix (by only dispatching rename update on `done` button). We could even dispatch on unfocus events potentially which might be more user friendly (not sure).

doctorn:
You can mark uncommitted changes as irrelevant actions and the existing machinery should handle this

doctorn:
See `homotopy_web::model::proof::Action::relevant`

calintat:
Generator renamings shouldn't be irrelevant though, because you want to be able to undo them. I think the best solution is to only update the name on done.

PR #648, opened 2022-08-02
Tweak rename history

thud:
This commit fixes an issue (#645) where multiple history events are
created for a single generator rename.

### DIFF
--- a/homotopy-web/src/app/signature/item.rs
+++ b/homotopy-web/src/app/signature/item.rs
@@ -269,7 +269,7 @@ impl ItemView {
                 SignatureItem::Item(info) => &info.name,
                 SignatureItem::Folder(name, _) => name,
             };
-            if self.name != prev_name.clone() {
+            if &self.name != prev_name {
                 apply_edit(
                     &ctx.props().dispatch,
                     ctx.props().node,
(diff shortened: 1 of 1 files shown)


homotopy-web/src/app/signature/item.rs
```diff
@@ -245,11 +258,26 @@ impl Component for ItemView {
...
             return false;
         }
 
+        // Apply rename if name has changed.
+        if mode == ItemViewMode::Viewing {
+            let prev_name = match &ctx.props().item {
+                SignatureItem::Item(info) => &info.name,
+                SignatureItem::Folder(name, _) => name,
+            };
+            if self.name != prev_name.clone() {
```
doctorn, line 272:
```rust
if &self.name != prev_name {
```

(I don't think you need to clone here)

homotopy-web/src/app/signature/item.rs
```diff
@@ -283,8 +311,8 @@ impl ItemView {
                 <input
                     type="text"
                     class="signature__item-name-input"
-                    value={name.clone()}
-                    oninput={ctx.link().callback(move |e: InputEvent| {
+                    value={self.name.clone()}
```
doctorn, line 314: `value={&self.name}` might work here too?

regular-citizen (approved):
Looks good to me

homotopy-web/src/app/signature/item.rs
thud: Fixed...

homotopy-web/src/app/signature/item.rs
thud: Yew doesn't like this for whatever reason. Annoying.

thud: Seems like Yew wants `&'static` references only for props.

PR #660, opened 2022-08-09
`Yew` keys for signature items and folders

thud:
This aims to fix a long standing issue with `Yew` improperly handling signature items and folders.


thud:
This would conflict quite severely with #632 so for now, it's based on top of the `signature-previews` branch. In future, we can either merge #632 (then this) into master or merge this PR into that branch first.

calintat:
This is amazing, thanks so much Jasper!!!

thud:
I'm gonna merge this into `signature-previews` and make it part of #632 for simplicity.

PR #664, opened 2022-08-10
Fix devserver reload for static files

### LINKED ISSUE #663: Devserver does not auto reload for static file edits
A fix should only require a relatively small change to `Makefile.toml`.

Steps to reproduce:
- `cargo make serve`
- edit eg. `styles.css` in `homotopy-web/static`

thud:
This fixes #663.


regular-citizen (approved):
Thanks Jasper, looks good!

PR #668, opened 2022-08-12
Improve GL color handling

thud:
- [x] Redefine gamma constant for more realistic render
- [x]  Make GL receive linear colors
- [x]  Make points, wires, surfaces differ in color from eachother (like svg)
- [x] Remove unnecessary normalisation operation (causes lack of precision)

### DIFF
--- a/homotopy-graphics/src/style.rs
+++ b/homotopy-graphics/src/style.rs
@@ -42,6 +42,10 @@ impl Color {
     {
         self.0.into_format().into_components()
     }
+
+    pub fn into_linear_f32_components(self) -> (f32, f32, f32) {
+        self.0.into_format::<f32>().into_linear().into_components()
+    }
 }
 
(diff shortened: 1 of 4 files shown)


doctorn (approved):
This is fantastic -- thanks so much

thud:
Great!

Any idea why the lint CI is failing for everything? Looks like a nix thing but idk how we would fix.

doctorn:
It looks like nix has jumped to Rust `1.63.0` which just released but some of the dependencies are pegged to an older compiler version. What happens if you delete `flake.lock` and then `nix build` locally?

doctorn:
(cc @NickHu because I'm making this up)

thud:
Nice, seems to work now. I'm gonna make it a separate PR for clarity and hopefully everything can work fine from now.

thud:
#669

PR #678, opened 2022-08-16
Resizeable drawer

thud:


H-Yulong:
I can see the resizeable drawer working! But they still look kind of wrong on my browser:



I think you might have to change the `overflow-y` attribute for `.drawer` class, it is currently set to `scroll`.



The "Signature" text at the top gets clipped and becomes "Sig...", which is not quite ideal. Maybe display it like "Sig." or don't display it at all would be better (if it's hard to implement then forget it)

calintat:
I think we shouldn't allow users to resize the drawer to less than its original width.

calintat:
Perhaps if the user tried to make the drawer smaller than its original width, we could automatically hide it.

thud:
Yeah, thanks. I was playing around with creating a semi-minimised view of the signature but I'm not sure if it's worth the effort. The logic for auto-closing the signature (if you resize it enough) is already implemented. It just needs some touching up.
Also still working on an expanded view for a more detailed view.

thud:
The `overflow-y: scroll` is necessary I think? If the user creates too many generators to show in one page.
Not completely sure but we might have to look at styling the scrollbar (and hiding when possible) if needed.

H-Yulong:
I think `overflow-y: auto` is exactly what we want

thud:
Ah cool.

calintat:
> Yeah, thanks. I was playing around with creating a semi-minimised view of the signature but I'm not sure if it's worth the effort. The logic for auto-closing the signature (if you resize it enough) is already implemented. It just needs some touching up.
> Also still working on an expanded view for a more detailed view.

That sounds great!

jamievicary:
We didn't discuss this today, can we please not leave pull requests hanging, is there any reason not to merge?

thud:
Quite right. Merging now.

PR #682, opened 2022-08-16
Better hovering mechanics for signature items

### LINKED ISSUE #679: Better hovering mechanics for signature items
Dimension should be showed in color sliver area (in left of signature item).
Settings cog should take its place when hovering.

thud:
This aims to improve hovering for generators in the signature. Specifically, it aims to stop the generator info from moving with every hover.

Closes #679.

### DIFF
--- a/homotopy-web/src/app/signature/item.rs
+++ b/homotopy-web/src/app/signature/item.rs
@@ -565,16 +565,7 @@ impl ItemView {
 
     fn view_info(&self, ctx: &Context<Self>) -> Html {
         match &ctx.props().item {
-            SignatureItem::Item(info) => {
-                html! {
-                    <>
-                        <span class="signature__item-child signature__generator-dimension">
-                            {info.diagram.dimension()}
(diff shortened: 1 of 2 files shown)


thud:
image

jamievicary:
Picture looks great jasper

jamievicary:
Maybe make the coloured box rounded on all 4 corners now.

thud:
Settings now experimentally only show when hovering over the settings/dimension box itself.
Does anyone have any opinions on this (easy to revert).

jamievicary:
Not sure I would need to try it out. The issue is that people may not realize the settings cog is there if it never occurs to them to hover their cursor over the dimension number.

calintat:
This looks fantastic, thanks so much Jasper!

I agree with Jamie that it's probably better to always show the settings icon when hovering.

calintat (approved):

thud:
Great, thanks. Merging now

PR #691, opened 2022-08-17
Fix performance regression with GL rendering

thud:
This moves code which is currently (unnecessarily) in `Renderer::render(..)` into `Scene::reload_meshes(..)` to fix the hit to performance which it provokes.

### DIFF
--- a/homotopy-web/src/app/diagram_gl/renderer/scene.rs
+++ b/homotopy-web/src/app/diagram_gl/renderer/scene.rs
@@ -1,28 +1,35 @@
-use std::mem;
+use std::{mem, rc::Rc};
 
 use homotopy_common::idx::IdxVec;
 use homotopy_core::{Diagram, Generator};
 use homotopy_graphics::{
     geom::{CubicalGeometry, SimplicialGeometry, VertData},
     gl::{array::VertexArray, GlCtx, Result},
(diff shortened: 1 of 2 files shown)


homotopy-web/src/app/diagram_gl/renderer/scene.rs
```diff
@@ -117,11 +126,11 @@ impl Scene {
         sphere_mesh.mk_point(p);
         sphere_mesh.inflate_point_3d(p, geometry_samples, &VertexShape::Circle);
         if let Some(sphere_buffers) = sphere_mesh.buffer_tris(ctx)?.into_iter().next() {
-            self.sphere = Some(vertex_array!(
+            self.sphere = Some(Rc::new(vertex_array!(
```
doctorn, line 129: Vertex arrays are reference counted under the hood so you don't need to add an `Rc` here, just `clone` the array

doctorn (changes requested):
Looks great + fixes the regression! Just one comment.

homotopy-web/src/app/diagram_gl/renderer/scene.rs
doctorn: Actually that's total nonsense ignore me this is correct

doctorn (approved):

PR #695, opened 2022-08-17
Migration tool

H-Yulong:
The migration tool for loading old homotopy.io files
It can:
- load the signature (name, id, colour, etc.) and the current workspace diagram

It cannot:
- load the slices of current workspace and the attaches (takes lots of efforts to implement, with little benefit)

### DIFF
--- a/homotopy-core/src/migration.rs
+++ b/homotopy-core/src/migration.rs
@@ -18,6 +18,8 @@ pub struct OldProof {
     #[serde(skip)]
     pub generator_info: Vec<OldGeneratorInfo>,
     #[serde(skip)]
+    pub workspace: Option<OldWorkspace>,
+    #[serde(skip)]
     generators: HashMap<usize, Generator>,
     #[serde(skip)]
     diagrams: HashMap<usize, Diagram>,
@@ -32,6 +34,12 @@ pub struct OldGeneratorInfo {
     pub name: String,
     pub color: String,
     pub diagram: Diagram,
+}
+
+pub struct OldWorkspace {
+    pub diagram: Diagram,
+    // pub path: Vec<SliceIndex>,
+    // pub view: u8,
 }
 
 impl OldProof {
@@ -55,7 +63,7 @@ impl OldProof {
             let i: usize = from_value(v["_l"].clone())?;
             self.load_generator(i)?;
         }
-        let workspace = self.load_workspace();
+        self.load_workspace()?;
         Ok(())
     }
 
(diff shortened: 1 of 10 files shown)


H-Yulong:
> How do I import a JSON file? I clicked on "Import" and it doesn't let me select JSON files.

There should be an option in the import window to select "all files"

calintat:
Right, that worked! But I think it will be confusing for users. Maybe we should change it so that the import accepts .hom files as well .json files:

https://github.com/homotopy-io/homotopy-rs/blob/e375617791f0175bc9751dd8924a6bc2d009c042/homotopy-web/src/app/project.rs#L33

calintat:
I managed to make it fail:
[homotopy.io - Wed Aug 17 2022 14 53 30 GMT+0100 (British Summer Time).json.zip](https://github.com/homotopy-io/homotopy-rs/files/9363258/homotopy.io.-.Wed.Aug.17.2022.14.53.30.GMT%2B0100.British.Summer.Time.json.zip)

Sorry I don't have a smaller example.

homotopy-web/src/model/migration.rs
```diff
@@ -0,0 +1,78 @@
...
+fn load(proof: OldProof) -> Option<(Signature, Option<Workspace>)> {
+    let mut signature: Signature = Default::default();
+
+    // generators gives a sorted list of GeneratorInfos
+    for v in proof.generator_info {
+        let color: Color = v.color.parse().ok()?;
+        let info = GeneratorInfo {
+            generator: v.generator,
+            name: v.name,
+            oriented: false,
```
calintat, line 55: Also can you please set this to true?

homotopy-core/src/migration.rs
```diff
@@ -0,0 +1,324 @@
...
+        let sublimits = self.generate_vec(slices_index)?;
+        let mut slices: Vec<Rewrite> = Vec::new();
+        for v in sublimits {
+            let i: usize = from_value(v["_l"].clone())?;
+            let c = self.load_rewrite(i)?;
+            slices.push(c);
+        }
+
+        let cone = Cone::new(cone_index, source, target, slices);
+        self.cones.insert(index, cone.clone());
```
calintat, line 261: Hmm what is `index` here? Cones are expected to be ordered by their index (i.e. `cone_index` above).

H-Yulong: That's for caching loaded cones, and `index` is the cone's array index in `stored`.

calintat: Makes sense

H-Yulong:
> Here is another one (still quite big I am afraid): [homotopy.io - Wed Aug 17 2022 15 00 56 GMT+0100 (British Summer Time).json.zip](https://github.com/homotopy-io/homotopy-rs/files/9363307/homotopy.io.-.Wed.Aug.17.2022.15.00.56.GMT%2B0100.British.Summer.Time.json.zip)

Added a fix, you can successfully import these files now

calintat:
Brilliant, thanks Yulong! Let's merge this then!

PR #704, opened 2022-08-18
make unselectable default in css

H-Yulong:
Everything is now un-selectable by default.
To make an element selectable, add "selectable" to its class.

### DIFF
--- a/homotopy-web/static/styles.css
+++ b/homotopy-web/static/styles.css
@@ -44,6 +44,9 @@ body, html {
   flex: 1;
   font-size: 14px;
   overflow: hidden;
+  -webkit-user-select: none; /* Safari */
+  -ms-user-select: none; /* IE 10 and IE 11 */
+  user-select: none; /* Standard syntax */
 }
 
 input, button {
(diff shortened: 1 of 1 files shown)


jamievicary:
Thanks for this, can you please ensure that the obvious things still remain selectable?

calintat:
Are there any UI elements we want to be selectable? I think the cell name is the only one and that will probably be automatically selectable when it's in edit mode.

jamievicary:
As per the original issue, at least the cell names and project metadata should be selectable.

H-Yulong:
The cell names are indeed selectable in the edit mode. But they're not selectable in normal mode, because when you click and drag (trying to select) on the cell name you're actually dragging the signature item (for reordering).

Also is project metadata in the master branch now? I left a reusable `selectable` css class so it won't be hard to make metadata selectable.

calintat:
Right, and I think 

> The cell names are indeed selectable in the edit mode. But they're not selectable in normal mode, because when you click and drag (trying to select) on the cell name you're actually dragging the signature item (for reordering).
> 
> Also is project metadata in the master branch now? I left a reusable `selectable` css class so it won't be hard to make metadata selectable.

The project metadata is not merged yet. I think it's best if the cell name is only selectable in edit mode. Thanks Yulong!

Issue #708, opened 2022-08-18
Using keybinding to select generators does not clear current workspace


H-Yulong:
With #572 fixed, the workspace is cleared when I select a generator with mouse, but not when I select with keybindings 1-9.

calintat:
Yes, I think I forgot to implement this for the `Select` action. Would you mind doing it Yulong?

PR #715, opened 2022-08-19
key select clears workspace

H-Yulong:
Solves #708.

### DIFF
--- a/homotopy-web/src/model/proof.rs
+++ b/homotopy-web/src/model/proof.rs
@@ -307,7 +307,7 @@ impl ProofState {
                         }
                     })
             }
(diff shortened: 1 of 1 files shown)


homotopy-web/src/model/proof.rs
```diff
@@ -307,7 +307,7 @@ impl ProofState {
                         }
                     })
             }
-            Action::Select(_) => self.workspace().map_or(true, |ws| ws.attach.is_some()),
+            Action::Select(_) => true,
```
calintat, line 310: You can combine this with the other actions that map to true.

calintat (approved):
Great, thanks Yulong! Just one comment.

PR #717, opened 2022-08-19
Colors for orientation information

### LINKED ISSUE #550: Orientation information in the renderer

thud:
Closes #550.

### DIFF
--- a/homotopy-graphics/src/style.rs
+++ b/homotopy-graphics/src/style.rs
@@ -36,6 +36,19 @@ impl Color {
         Self(palette::Lighten::lighten(self.0.into_linear(), amount).into())
     }
 
+    // we combine saturate and lighten into a single function since we need to convert SRGB into
+    // HSL or HSV color space anyway and we save some computation by lightening this directly
+    // instead of the resulting SRGB.
+    #[must_use]
+    pub fn desaturate_and_lighten(&self, desaturate: f32, lighten: f32) -> Self {
+        let hsl: palette::Hsl =
+            palette::convert::FromColor::from_color(self.0.into_format::<f32>());
+        let desaturated = palette::Saturate::saturate(hsl, -desaturate);
+        let lightened = palette::Lighten::lighten(desaturated, lighten);
+        let srgb: palette::Srgb<f32> = palette::convert::FromColor::from_color(lightened);
+        Self(srgb.into_format())
+    }
+
     pub fn is_light(&self) -> bool {
         palette::RelativeContrast::get_contrast_ratio(
             palette::Srgb::new(1., 1., 1.),
@@ -60,7 +73,9 @@ impl FromStr for Color {
     type Err = palette::rgb::FromHexError;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        palette::Srgb::<u8>::from_str(s).map(Self)
+        palette::Srgb::<u8>::from_str(s)
+            .map(palette::Srgb::into_format)
+            .map(Self)
     }
 }
 
--- a/homotopy-graphics/src/style.rs
+++ b/homotopy-graphics/src/style.rs
@@ -71,8 +71,24 @@ impl Color {
         ) < 1.5
     }
 
-    // Get color in lightened form (based on offset [0, 8]). We wrap colors so that they are
-    // sufficiently different and clearly distinguishable for end users.
+    // Get color in lightened form (based on offset [0, 8]). We wrap colors if needed so that they
+    // are sufficiently different and clearly distinguishable for end users.
+    //
+    // Colours are calculated from C and R where:
+    //      C = max(0, D - N - K)
+    //      R = -1 | 0 | 1          // orientation of generator
+    //      D = 0...                // diagram dimension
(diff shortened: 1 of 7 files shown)


thud:
This is still early in prototype phase, trialling different ways of encoding orientation in the colours of rendered generators.

I have also used this opportunity to switch the internal scalar for colors to `T = f32` from `T = u8`, which should give slightly more accurate colour operations and avoids loss of precision for `gl` rendering etc.

thud:
In hindsight, it's likely not a good idea to make the change to `T = f32`, given that it will break the serialisation code and therefore imports for previously generated diagrams/signatures.

calintat:
> In hindsight, it's likely not a good idea to make the change to `T = f32`, given that it will break the serialisation code and therefore imports for previously generated diagrams/signatures.

Not a problem, I literally broke serialization on the labelled branch this morning :)

jamievicary:
Looking great! But we need to tweak the values I think. What have you used at the moment for the 9 values, can you list the nine saturation/lightness values you have used?

calintat:
Thanks Jasper! We actually need three different saturation levels per generator instead of two. These correspond to the three different orientations (-1, 0, 1). So there will be 9 CSS classes in total.

calintat:
> Looking great! But we need to tweak the values I think. What have you used at the moment for the 9 values, can you list the nine saturation/lightness values you have used?

I agree! I can literally not tell them apart right now (granted I am colourblind). Honestly I still think arrows and different shapes would be a better way to tell inverses apart than colours.

thud:
Yeah, I personally agree about having arrows and boxes but (especially for 3d) colours is a good option to have.
All the colour modifications are made in `homotopy-graphics/src/svg.rs` so for exact numbers look there.

The saturation factors as of now are:
| orientation | saturation |
|----------|------|
| forward  | 0    |
| zero?    |      |
| inverted | -0.3 |

The lightening factors as of now (unchanged) are:
| codimension | lightness |
|----------|------|
| point   | 0    |
| wire    | 0.05 |
| surface | 0.1  |

The factors are applied by `palette` (documented [here](https://ogeon.github.io/docs/palette/master/palette/trait.Saturate.html) for saturation).

I think we should also consider using `palette` from the `crates` registry instead of the `github` because they seem to keep changing their API (at least in the docs).

jamievicary:
Hi Jasper, I cannot see yet how this implements the 9 saturation/lightness values described in issue #550, can you give more detail please.

doctorn:
> In hindsight, it's likely not a good idea to make the change to `T = f32`, given that it will break the serialisation code and therefore imports for previously generated diagrams/signatures.

Sounds like time for some `obake` antics 🤩

calintat:
Should we try -0.25 for zero and -0.5 for inverses? We can tweak these later.

jamievicary:
I'm making further comments in the issue #550

thud:
@jamievicary, I've added a temporary set of sliders in the settings panel for quickly setting lightnesses. This does not yet implement 'fixed' colour lightness modifications (with wrapping etc.).

jamievicary:
Excellent, do these serialise with the workspace?

thud:
Not at the moment, but they could do if needed?

calintat:
The slider of inverse wire lightness doesn't seem to be working.

thud:
> The slider of inverse wire lightness doesn't seem to be working.
Yup, fixed now.

thud:
Todo:
- [ ] Colours for tikz, manim, gl etc.
- [ ] Colours should not change when entering slices.

thud:
Yup ok. Do colours look right now? I think this is accurate to Jamie's formula.

calintat:
I tried doing a snake and the inverse wire looks almost white:

homotopy_io_export-2

Honestly I think this idea is fundamentally flawed, it's too much information to encode using only colours.

thud:
Hmm yeah, I'll see if I can quickly get a better colour solution prototyped.

thud:
Alright, I have a solution which works pretty well for the time being (all the preset colours look alright in different orientations though the black and white presets have similar looking inverses). Unless someone is unhappy with this change, I'll get this merged very soon.

I've added a note which explains how colours are calculated in the most recent commit.

https://github.com/homotopy-io/homotopy-rs/blob/66febb5190b0005df2e67c0a2e8b933a734f9629/homotopy-graphics/src/style.rs#L57-L74

thud:
Alright, I've implemented colours as decided by Jamie. This is to be merged imminently.

https://github.com/homotopy-io/homotopy-rs/blob/1c6f8b2f894915b02e0b795bfde2115371fcf4e1/homotopy-graphics/src/style.rs#L51-L76

jamievicary:
Thanks Jasper this is great. Next can you please make sure the colours are being correctly displayed in 3d and 4d please.

PR #725, opened 2022-08-21
add serde default to metadata fields

H-Yulong:
Now the migration tool can load projects with some of the metadata fields missing.

### DIFF
--- a/homotopy-web/src/model/migration.rs
+++ b/homotopy-web/src/model/migration.rs
@@ -14,10 +14,13 @@ struct Export {
 
 #[derive(Deserialize)]
 struct Metadata {
+    #[serde(default)]
     #[serde(rename = "title")]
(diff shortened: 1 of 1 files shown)


calintat:
Interesting. How come the workspace could be migrated if you clear the diagram?

H-Yulong:
Interesting indeed, that's a super-legacy issue I suspect. Jamie's proof file might be exported from some really ancient version of homotopy-js, which does not have `abstract` field in the metadata. When you clear the diagram and export it (from the current version of homotopy-js), it adds the abstract field to the proof file, hence solving the problem. If you try import the original file to homotopy.io, do nothing, then export, then the migration tool can load this new file correctly.

calintat:
I see, that makes sense. Let's merge this then!

Issue #539, opened 2022-06-20
UI testing


NickHu:
The architecture of the program is well-suited for unit testing (every UI state can be reached by an appropriate sequence of `Action`s, which is extremely easy to encode versus something like "click on this part, wait for it to load, activate this button"). In principle, it should be easy to write unit tests to ensure that it behaves nicely using this approach, and we should, but it's not something that we do currently.

NickHu:
@zrho knows more about this, anything to add?

PR #748, opened 2022-08-25
Homotopy-model subcrates

regular-citizen:
This PR introduces two new subcrates: homotopy-model and homotopy-cli.

Homotopy-model mostly contains code that used to be in homotopy-web/src/model, but has been isolated to compile to native binaries.

Homotopy-cli imports homotopy-model and builds a binary that should give us access to large parts of the codebase within a native binary, with the aim to improve overall testability and debuggabilty of the project. For example, things like #539 become realistic.


calintat:
Thanks Chiara, this is looking very promising!

I have thought about it more, and I think it's better to have homotopy-model depend on homotopy-graphics rather than the other way around. I'm just really uncomfortable with having `style.rs` in model, because it really belongs in graphics.

calintat:
And I guess if we allow model to depend on graphics, we also gain access to the SVG/TikZ exports, so the CLI can live inside model and it doesn't need to be a separate crate.

regular-citizen:
> I have thought about it more, and I think it's better to have homotopy-model depend on homotopy-graphics rather than the other way around. I'm just really uncomfortable with having `style.rs` in model, because it really belongs in graphics.
> And I guess if we allow model to depend on graphics, we also gain access to the SVG/TikZ exports, so the CLI can live inside model and it doesn't need to be a separate crate.

Thanks for the feedback, I like your action plan.
I'll try to move gl stuff out of graphics and into web, remove wasm deps in graphics and change the dependency graph, reverting the style.rs move.

calintat:
Yes, I agree that graphics shouldn't depend on wasm. I think everything in homotopy-graphics/gl can go to homotopy-web/app/diagram_gl. After you do that, I might try to clean up the structure of graphics a bit.

regular-citizen:
> Yes, I agree that graphics shouldn't depend on wasm. I think everything in homotopy-graphics/gl can go to homotopy-web/app/diagram_gl. After you do that, I might try to clean up the structure of graphics a bit.

Do you want it to be in diagram_gl or in diagram_gl/gl or some other variation? Right now I'm just making a homotopy-web/src/gl folder, but it feels wrong.

calintat:
Hmm it should be under diagram_gl, but I'm not sure if we should mix it with the stuff in diagram_gl/renderer or have a different subfolder (I don't really like the name diagram_gl/gl haha). @doctorn any thoughts since this is your code??

regular-citizen:
> Hmm it should be under diagram_gl, but I'm not sure if we should mix it with the stuff in diagram_gl/renderer or have a different subfolder (I don't really like the name diagram_gl/gl haha). @doctorn any thoughts since this is your code??

I was thinking of doing diagram_gl/gl for the old homotopy-graphics/src/gl, and move gl.rs and buffers.rs to diagram_gl.
I think it's not too bad as an alternative, and keeping the submodule just "gl" minimises the diffs for all the things that are importing "gl::*".

calintat:
Do that for now and also let's see what Nathan says.

doctorn:
Could we have a crate dedicated to the `src/gl` component of `homotopy_graphics`? It's a completely standalone module that implements a wrapper around WebGL. I was planning to make a seperate crate out of it anyway at some point... Moving it into the application code would work but it opens us up to accidentally losing the decoupling we have at the moment which is quite a nice property!

regular-citizen:
> Could we have a crate dedicated to the `src/gl` component of `homotopy_graphics`? It's a completely standalone module that implements a wrapper around WebGL. I was planning to make a seperate crate out of it anyway at some point... Moving it into the application code would work but it opens us up to accidentally losing the decoupling we have at the moment which is quite a nice property!

Makes sense, I'll go for this.

doctorn:
> 



> > Could we have a crate dedicated to the `src/gl` component of `homotopy_graphics`? It's a completely standalone module that implements a wrapper around WebGL. I was planning to make a seperate crate out of it anyway at some point... Moving it into the application code would work but it opens us up to accidentally losing the decoupling we have at the moment which is quite a nice property!
> 
> Makes sense, I'll go for this.

If you do something like
```
[dependencies]
gl = { package = "homotopy_webgl", path = "../homotopy-webgl" }
```
the `gl::` import head will stay the same

regular-citizen:
> > 
> 
> > > Could we have a crate dedicated to the `src/gl` component of `homotopy_graphics`? It's a completely standalone module that implements a wrapper around WebGL. I was planning to make a seperate crate out of it anyway at some point... Moving it into the application code would work but it opens us up to accidentally losing the decoupling we have at the moment which is quite a nice property!
> > 
> > 
> > Makes sense, I'll go for this.
> 
> If you do something like
> 
> ```
> [dependencies]
> gl = { package = "homotopy_webgl", path = "../homotopy-webgl" }
> ```
> 
> the `gl::` import head will stay the same

It is done, homotopy-gl is born!

PR #749, opened 2022-08-25
Image export

H-Yulong:
Partially solves issue #525.

### DIFF
--- a/homotopy-web/src/app/image_export.rs
+++ b/homotopy-web/src/app/image_export.rs
@@ -1,0 +1,75 @@
+use yew::prelude::*;
+
+use crate::model::proof::{Action, AttachOption, Signature};
+
+#[derive(Debug, Clone, PartialEq, Properties)]
+pub struct Props {
+    pub dispatch: Callback<Action>,
+    pub options: im::Vector<AttachOption>,
+    pub signature: Signature,
+}
+
+#[derive(Debug, Clone)]
+pub enum Message {}
+
+pub struct ImageExportView;
+
+impl Component for ImageExportView {
+    type Message = Message;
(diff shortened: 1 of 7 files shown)


homotopy-web/src/app/sidebar.rs
```diff
@@ -156,9 +156,27 @@ impl Component for Sidebar {
...
-        let attach_options = ctx
-            .props()
-            .proof
+        let proof = &ctx.props().proof;
+
+        if proof.show_image_export {
+            return html! {
+                <SidebarDrawer
+                    // yh: Re-using the attach style class here
+                    class="attach"
```
calintat, line 165: Hmm could we rename this to something other than attach. Maybe "dialog" because these things behave like dialogs?

calintat:
This looks very good Yulong!

I don't remember if we decided to merge this now or after you add the options, but feel free to do whatever you think is best. In terms of options, let's start with a toggle "Show masking" under the TikZ export. You can ignore the value for now.

calintat:
Chiara says merge now actually.

homotopy-web/src/app/image_export.rs
```diff
@@ -0,0 +1,97 @@
...
+        }
+    }
+}
+
+impl ImageExportView {
+    fn view_tikz(ctx: &Context<Self>) -> Html {
+        if ctx.props().view_dim == 2 {
+            html! {
+                <>
+                    <h3>{"Export to Tikz"}</h3>
```
calintat, line 50: This should be **TikZ** with a capital Z.

H-Yulong:
Cool. I think it'd be better not to close #525 now, since there are options to be done.

PR #750, opened 2022-08-25
Action replay

regular-citizen:
This PR adds a method to dump actions from the web interface, and support for replays in the CLI.

### DIFF
--- a/homotopy-web/src/app/project.rs
+++ b/homotopy-web/src/app/project.rs
@@ -35,6 +35,10 @@ impl Component for ProjectView {
 
     fn view(&self, ctx: &Context<Self>) -> Html {
         let export = ctx.props().dispatch.reform(|_| model::Action::ExportProof);
+        let actions = ctx
+            .props()
+            .dispatch
+            .reform(|_| model::Action::ExportActions);
         let import = ctx.link().callback(|e: Event| {
             let input: HtmlInputElement = e.target_unchecked_into();
             if let Some(filelist) = input.files() {
@@ -50,6 +54,7 @@ impl Component for ProjectView {
                     {"Import"}
                 </label>
                 <input type="file" accept="application/msgpack,.hom,.json" class="visually-hidden" id="import" onchange={import}/>
+                <button onclick={actions}>{"Export Actions"}</button>
                 <input
                     type="text"
                     class="metadata_title"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -42,6 +42,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "arbitrary"
+version = "1.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5a7924531f38b1970ff630f03eb20a2fde69db5c590c93b0f3482e95dcc5fd60"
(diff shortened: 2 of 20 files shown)


homotopy-web/src/app/project.rs
```diff
@@ -50,6 +54,7 @@ impl Component for ProjectView {
                     {"Import"}
                 </label>
                 <input type="file" accept="application/msgpack,.hom,.json" class="visually-hidden" id="import" onchange={import}/>
+                <button onclick={actions}>{"Export Actions"}</button>
```
doctorn, line 57: I'm wondering if this should be in our debug menu?

doctorn (commented):
You have no idea how happy this makes me

homotopy-web/src/app/project.rs
regular-citizen: It most definitely should. However, I''m a webdev dumb dumb and couldn't figure it out, so any help would be appreciated on this.

regular-citizen: Thanks :)

calintat: Haha I also said it should be in the debug panel. My only worry is that the debug is not available on the live preview which is slightly annoying.

Btw Chiara does the normal export contain the action history? If someone reports a bug and attaches the workspace export, we should be able to get the actions from that.

regular-citizen: It doesn't, but I agree it should. I think what it comes down to is that one of the data structures relating to the proof tree contains a timestamp of type Instant, which is not serialiser-friendly.

calintat: Right, that's problematic because in practice it's unlikely people will construct a large diagram in one go without saving and re-loading (which would destroy the action history).

regular-citizen: Do you know why we are timestamping everything so aggressively? Several bits of code mention "determinism" but I have no idea what that has to do with anything.

regular-citizen: @doctorn ok so I gave another shot at moving the button and I can't figure out how to make the action dispatch happen. If anyone else wants to give it a shot, feel free to implement it, but I don't want to spend the whole day on it.

doctorn: > Do you know why we are timestamping everything so aggressively? Several bits of code mention "determinism" but I have no idea what that has to do with anything.

(CC @NickHu)

doctorn: > @doctorn ok so I gave another shot at moving the button and I can't figure out how to make the action dispatch happen. If anyone else wants to give it a shot, feel free to implement it, but I don't want to spend the whole day on it.

I can have a look after work?

regular-citizen:
Ready to merge now, we'll handle serializing actions into .homs in a different PR. I have more patches to merge and I'd rather base them off master than on this.

homotopy-web/src/app/project.rs
regular-citizen: Sure, I'd really appreciate it. I will merge this in the meantime tho.

PR #752, opened 2022-08-26
Panic-proofing the tool

regular-citizen:
Having isolated model into its own crate, and having made all the relevant traits compatible with Arbitrary, we can now use a fuzzer to construct action sequences that lead to panics.

This should vastly reduce the ways in which the tool can crash on users.

### DIFF
--- a/homotopy-common/src/tree.rs
+++ b/homotopy-common/src/tree.rs
@@ -9,7 +9,6 @@ use crate::{declare_idx, idx::IdxVec};
 
 declare_idx! {
     #[derive(serde::Serialize, serde::Deserialize)]
-    #[cfg_attr(feature = "fuzz", derive(arbitrary::Arbitrary))]
     pub struct Node = usize;
 }
 
--- a/homotopy-common/src/tree.rs
+++ b/homotopy-common/src/tree.rs
@@ -226,7 +226,7 @@ impl<T> Tree<T> {
         // Don't introduce a cycle
         assert!(self.ancestors_of(parent).all(|ancestor| ancestor != node));
 
-        if node.0 >= self.nodes.len() {
+        if node.0 >= self.nodes.len() || parent.0 >= self.nodes.len() {
             return;
         }
 
--- a/homotopy-common/src/tree.rs
+++ b/homotopy-common/src/tree.rs
@@ -120,19 +120,27 @@ impl<T> Tree<T> {
     }
 
     #[inline]
-    pub fn with<F, U>(&self, node: Node, f: F) -> U
+    pub fn with<F, U>(&self, node: Node, f: F) -> Option<U>
     where
         F: FnOnce(&NodeData<T>) -> U,
     {
-        f(&self.nodes[node])
+        if node.0 < self.nodes.len() {
+            Some(f(&self.nodes[node]))
+        } else {
+            None
+        }
     }
 
     #[inline]
-    pub fn with_mut<F, U>(&mut self, node: Node, f: F) -> U
+    pub fn with_mut<F, U>(&mut self, node: Node, f: F) -> Option<U>
     where
         F: FnOnce(&mut NodeData<T>) -> U,
     {
-        f(&mut self.nodes[node])
+        if node.0 < self.nodes.len() {
+            Some(f(&mut self.nodes[node]))
+        } else {
+            None
+        }
     }
 
     /// Removes `node` from the tree. This is done by disconnecting the subtree rooted at `node`
@@ -143,22 +151,27 @@ impl<T> Tree<T> {
     /// Doing so will free the memory associated with all disconnected components.
     #[inline]
     pub fn remove(&mut self, node: Node) {
-        if let Some(parent) = self.nodes[node].parent {
-            self.nodes[parent].children.retain(|child| *child != node);
+        if node.0 < self.nodes.len() {
+            if let Some(parent) = self.nodes[node].parent {
+                self.nodes[parent].children.retain(|child| *child != node);
+            }
+            self.nodes[node].parent = None;
         }
-
-        self.nodes[node].parent = None;
     }
 
     #[inline]
-    pub fn push_onto(&mut self, node: Node, t: T) -> Node {
-        let id = self.nodes.push(NodeData {
-            data: t,
-            children: vec![],
-            parent: Some(node),
-        });
-        self.nodes[node].children.push(id);
-        id
+    pub fn push_onto(&mut self, node: Node, t: T) -> Option<Node> {
+        if node.0 < self.nodes.len() {
+            let id = self.nodes.push(NodeData {
+                data: t,
+                children: vec![],
+                parent: Some(node),
+            });
+            self.nodes[node].children.push(id);
+            Some(id)
+        } else {
+            None
+        }
     }
 
     /// Returns an iterator of the ancestors of `node`.
@@ -214,6 +227,10 @@ impl<T> Tree<T> {
         // Don't introduce a cycle
         assert!(self.ancestors_of(parent).all(|ancestor| ancestor != node));
 
+        if node.0 >= self.nodes.len() {
+            return;
+        }
+
         if let Some(old_parent) = self.nodes[node].parent {
             self.nodes[old_parent]
                 .children
@@ -239,6 +256,10 @@ impl<T> Tree<T> {
 
         // Fast return when we're trying to reparent a node next to itself
         if node == successor {
+            return;
+        }
+
+        if successor.0 >= self.nodes.len() {
             return;
         }
 
@@ -378,7 +399,7 @@ impl<'a, T> Iterator for AncestorIterator<'a, T> {
     #[inline]
     fn next(&mut self) -> Option<Node> {
         let node = self.current?;
-        self.current = self.tree.with(node, NodeData::parent);
+        self.current = self.tree.with(node, NodeData::parent)?;
         Some(node)
     }
 }
--- a/homotopy-core/src/attach.rs
+++ b/homotopy-core/src/attach.rs
@@ -1,6 +1,5 @@
 use std::convert::{From, Into, TryInto};
 
-use either::Either;
 use serde::{Deserialize, Serialize};
 
 use crate::{
@@ -60,13 +59,10 @@ impl From<Boundary> for BoundaryPath {
     }
 }
 
-pub fn attach<F, E>(
-    diagram: &DiagramN,
-    path: BoundaryPath,
-    build: F,
-) -> Result<DiagramN, Either<E, DimensionError>>
+pub fn attach<F, E>(diagram: &DiagramN, path: BoundaryPath, build: F) -> Result<DiagramN, E>
 where
     F: FnOnce(Diagram) -> Result<Vec<Cospan>, E>,
+    E: From<DimensionError>,
 {
     let (diagram, _) = attach_worker(diagram, path, build)?;
     Ok(diagram)
@@ -76,13 +72,14 @@ fn attach_worker<F, E>(
     diagram: &DiagramN,
     path: BoundaryPath,
     build: F,
-) -> Result<(DiagramN, usize), Either<E, DimensionError>>
+) -> Result<(DiagramN, usize), E>
 where
     F: FnOnce(Diagram) -> Result<Vec<Cospan>, E>,
+    E: From<DimensionError>,
 {
     match path {
         BoundaryPath(Boundary::Source, 0) => {
-            let mut cospans = build(diagram.source()).map_err(Either::Left)?;
+            let mut cospans = build(diagram.source())?;
             let mut source = diagram.source();
 
             for cospan in cospans.iter().rev() {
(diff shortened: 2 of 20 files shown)


jamievicary:
Can you spell out the consequences of this for me a bit please Chiara. Surely if there are bugs in the code causing crashes a change like this can't provide some sort of generic fix.

Also would this allow us to emit, when the tool crashes, an automatic trace of recent activity that can be imported by someone else to reproduce the crash?

regular-citizen:
> Can you spell out the consequences of this for me a bit please Chiara. Surely if there are bugs in the code causing crashes a change like this can't provide some sort of generic fix.

You're correct, this is not a generic fix against bugs in algorithms. What the changes in this PR are meant to do is put us in a situation where algorithms may fail or return the wrong results in some circumstances, but will never/rarely cause panics.

Right now most of the bugs the fuzzer is exposing are nearly impossible to trigger big hand.
However, since the randomized traces produced by the fuzzer are so outlandishly far from what a human would do, they help shine a light on where/how we're failing to handle panic-able errors properly.
For example, I found out that our code that checks "validity of some actions" gave automatic chart-blanche to everything to do with modifying the signature, no matter how nonsensical.

Another nice side effect of squashing these low-hanging panics is that the fuzzer starts producing more interesting traces. I'm at the point where I'm starting to see contraction/expansion invocations show up in randomized traces.

> Also would this allow us to emit, when the tool crashes, an automatic trace of recent activity that can be imported by someone else to reproduce the crash?

Yes, that is something we could in theory already support on top of the merged code in homotopy-model. However, things like these are much easier to do if the code isn't completely dead due to panics, as panics are the least actionable form of error.

homotopy-core/src/common.rs
```diff
@@ -51,12 +51,20 @@ pub type SingularHeight = usize;
...
 
 #[derive(PartialEq, Eq, Copy, Clone, Debug, Hash)]
-#[cfg_attr(feature = "fuzz", derive(arbitrary::Arbitrary))]
 pub enum Height {
     Singular(SingularHeight),
     Regular(RegularHeight),
 }
 
+#[cfg(feature = "fuzz")]
+impl<'a> arbitrary::Arbitrary<'a> for Height {
```
calintat, line 60: Can't this be derived?

homotopy-core/src/attach.rs
```diff
@@ -71,13 +76,13 @@ fn attach_worker<F, E>(
...
     build: F,
-) -> Result<(DiagramN, usize), E>
+) -> Result<(DiagramN, usize), Either<E, DimensionError>>
 where
     F: FnOnce(Diagram) -> Result<Vec<Cospan>, E>,
 {
     match path {
         BoundaryPath(Boundary::Source, 0) => {
-            let mut cospans = build(diagram.source())?;
+            let mut cospans = build(diagram.source()).map_err(|e| Either::Left(e))?;
```
calintat, line 85: You can just do `map_err(Either::Left)`.

homotopy-core/src/attach.rs
```diff
@@ -91,15 +96,15 @@ where
         }
 
         BoundaryPath(Boundary::Target, 0) => {
-            let added_cospans = build(diagram.target())?;
+            let added_cospans = build(diagram.target()).map_err(|e| Either::Left(e))?;
```
calintat, line 99: See above.

homotopy-core/src/attach.rs
```diff
@@ -91,15 +96,15 @@ where
...
             let offset = added_cospans.len();
             let mut cospans = diagram.cospans().to_vec();
             cospans.extend(added_cospans);
             Ok((DiagramN::new(diagram.source(), cospans), offset))
         }
 
         BoundaryPath(boundary, depth) => {
-            let source: DiagramN = diagram.source().try_into().unwrap();
+            let source: DiagramN = diagram.source().try_into().map_err(|e| Either::Right(e))?;
```
calintat, line 107: See above.

homotopy-core/src/common.rs
regular-citizen: Yes, and it originally was. I changed it to avoid having to spend more time debugging overflow errors which are likely to never show up in reality. For example, when we convert from the Singular/Regular heights indexing into usize, we multiple by 2, which can in theory cause overflow, but it's never actually gonna be a problem and also it's most definitely not work fixing everywhere.

The value I set here is high enough to be able to cause out of bounds error wherever they haven't been fixed.

homotopy-model/src/model/proof.rs
```diff
@@ -287,7 +287,11 @@ impl ProofState {
                     selected.boundary == boundary || globularity(&selected.diagram, &ws.diagram)
                 })
             }),
-            Action::TakeIdentityDiagram | Action::ClearWorkspace => self.workspace.is_some(),
+            Action::TakeIdentityDiagram => self
+                .workspace
+                .as_ref()
+                .map_or(false, |w| w.view.dimension < u8::MAX),
```
calintat, line 293: Why are you checking the view dimension? The `TakeIdentityDiagram` action is valid regardless of the view dimension (it only increases the view dimension if it's 1D).

homotopy-common/src/tree.rs
```diff
@@ -120,19 +119,27 @@ impl<T> Tree<T> {
     }
 
     #[inline]
-    pub fn with<F, U>(&self, node: Node, f: F) -> U
+    pub fn with<F, U>(&self, node: Node, f: F) -> Option<U>
     where
         F: FnOnce(&NodeData<T>) -> U,
     {
-        f(&self.nodes[node])
+        if node.0 < self.nodes.len() {
```
calintat, line 126: Guarding against index out of bounds errors feel a bit unnecessary to me.

homotopy-model/src/model/proof.rs
regular-citizen: This check actually belongs to UpdateView, you're right. It was mean to prevent overflow panics when the dimension is 255.

homotopy-common/src/tree.rs
regular-citizen: Yeah those maybe were a bit excessive, but plenty of other places were calling into it without doing any bounds checking.

homotopy-core/src/contraction.rs
```diff
@@ -108,8 +108,9 @@ impl DiagramN {
                 signature,
             )?;
 
-            Ok(vec![cospan])
+            Ok::<_, ContractionError>(vec![cospan])
```
calintat, line 111: Hmm I really don't like that we have to use an `Either` because of this. I think `ContractionError` should contain `DimensionError` but not sure how to change the type of `attach` to only allow `E` types which contain `DimensionError` as a subtype.

regular-citizen: I'm quite unhappy with the either dependency as well. Can thiserror handle these kinds of things?

homotopy-core/src/contraction.rs
```diff
@@ -108,8 +108,9 @@ impl DiagramN {
...
         })
+        .or(Err(ContractionError::Invalid))
```
calintat, line 113: It feels very wrong to me to use `ContractionError::Invalid` here. This suggests the contraction was invalid (as in things didn't contract).

homotopy-core/src/expansion.rs
```diff
@@ -68,8 +68,9 @@ impl DiagramN {
 
             typecheck_cospan(slice, cospan.clone(), boundary_path.boundary(), signature)?;
 
-            Ok(vec![cospan])
+            Ok::<_, ExpansionError>(vec![cospan])
         })
+        .or(Err(ExpansionError::NoComponent))
```
calintat, line 73: Same comment for contraction. `NoComponent` means something else.

calintat:
This looks good, thanks Chiara!

My only concern is that you were a bit too aggressive in wrapping things in `Result` / `Option` and propagating errors up. Sometimes we don't want to propagate errors up. For example, in contraction propagating the attach error feels wrong because contraction errors are shown to the user as toasts and the user won't be able to tell the difference between "contraction didn't go through" and "something was not implemented correctly" errors. Panicking might actually be preferable in the latter case, because it would encourage the user to make a bug report.

homotopy-model/src/model/proof.rs
calintat: Even for `UpdateView` you shouldn't need to check anything, right? You can't construct `View`s which are not between 0 and 4.

regular-citizen:
Very fair, I had a sense I was wrapping things a bit too enthusiastically. I see two ways forward for dealing with the specific instances of unwanted wrapping:
- Either we ".expect" on the spot, and give move descriptive panic messages as to why something should not have happened,
- We bubble up an AlgorithmErrorContactDevelopers variant, to be shown in a toaster.

I prefer the second option, but both are better than unwrap hell.

homotopy-model/src/model/proof.rs
regular-citizen: Then I would propose to make it explicit in the "UpdateView" is_valid criterion. Right now update_view just saves whatever you pass it.

calintat:
> Very fair, I had a sense I was wrapping things a bit too enthusiastically. I see two ways forward for dealing with the specific instances of unwanted wrapping:
> 
> * Either we ".expect" on the spot, and give move descriptive panic messages as to why something should not have happened,
> * We bubble up an AlgorithmErrorContactDevelopers variant, to be shown in a toaster.
> 
> I prefer the second option, but both are better than unwrap hell.

I'm sorry but I am really not in favour of the second option. More descriptive panic messages would be nice I guess.

regular-citizen:
Cool, I'll implement it. It contract/expansion the places you want this in the most?

homotopy-core/src/contraction.rs
```diff
@@ -108,9 +108,12 @@ impl DiagramN {
...
             )?;
 
-            Ok::<_, ContractionError>(vec![cospan])
-        })
-        .or(Err(ContractionError::Invalid))
+            Ok(vec![cospan])
+        }) {
+            Ok(d) => Ok(d),
+            Err(either::Either::Left(e)) => Err(e),
+            Err(either::Either::Right(e)) => panic!("Attach in contract failed due to a dimension error. Please open an issue on GitHub with the action dump. Error: {}", e),
```
calintat, line 115: That's a bit too verbose and it feels weird to tell people to report this on GitHub in the panic message. I would go with something like "boundary path in contraction was invalid".

homotopy-core/src/expansion.rs
```diff
@@ -68,9 +68,12 @@ impl DiagramN {
...
             typecheck_cospan(slice, cospan.clone(), boundary_path.boundary(), signature)?;
 
-            Ok::<_, ExpansionError>(vec![cospan])
-        })
-        .or(Err(ExpansionError::NoComponent))
+            Ok(vec![cospan])
+        }) {
+            Ok(d) => Ok(d),
+            Err(either::Either::Left(e)) => Err(e),
+            Err(either::Either::Right(e)) => panic!("Attach in expand failed due to a dimension error. Please open an issue on GitHub with the action dump. Error: {}", e),
```
calintat, line 75: Same here: "boundary path in expansion was invalid"

homotopy-model/src/model/proof.rs
```diff
@@ -327,6 +323,7 @@ impl ProofState {
                 .workspace
                 .as_ref()
                 .map_or(false, |ws| ws.visible_dimension() > 0),
+            Action::UpdateView(v) => v.dimension <= 4,
```
calintat, line 326: I don't understand why we need this. Views are constrained to 0-4 (see their constructor).

calintat: Actually this reminds me that I would like View to be an enum. I don't remember why I never did that.

homotopy-core/src/contraction.rs
calintat: I have an idea. We can add `DimensionError` to `ContractionError` like this:

```rust
struct ContractionError {
   InvalidBoundaryPath(#[from] DimensionError)
}
```

and that automatically derives `impl From<DimensionError> for ContractionError`. Then we need to change the type of `attach` to add the constraint `E : From<DimensionError>`. That way `attach` can just return `E` instead of an either type.

calintat: And you won't even need to do any conversions manually because `?` handles all of that for you.

calintat: One more thing. If you look at `DiagramN::attach`:
https://github.com/homotopy-io/homotopy-rs/blob/ab001c9b512c4f5e352b0d72ded368b856954992/homotopy-core/src/diagram.rs#L450-L469

That already throws a dimension error if the boundary path is invalid. Since attach now handles that, we can remove that check and unify `AttachmentError::Dimension` and `DimensionError`.

calintat: I did this. I am now happy.

calintat (approved):

regular-citizen:
Thanks Calin for this thorough review, I know this patch was kind of a lot. I'm pushing the final two liner change to remove the view stuff and then I'll merge.

NickHu:
We should in general use `expect` a bit more over `unwrap` probably. I'm not sure if it's considered idiomatic rust though.

PR #755, opened 2022-08-27
Debug renderer

NickHu:
Far too frequently we need to get into the nitty gritty of a diagram, and this involves lots of tedious drawing by hand when we have a perfectly good 3D renderer already!

This is my attempt to add a function that can be called **anywhere within homotopy-core**, which when called will hijack the app and display a rotatable debug model of the diagram passed in:
image

It's a rough prototype, and I definitely need some help with the graphics stuff, but it's a decent way there.

Note that this is a debugging tool, so there is nothing to play with in the release version that the PR preview generates. To test it out, apply this patch:

```diff
diff --git a/homotopy-core/src/diagram.rs b/homotopy-core/src/diagram.rs
index 8e8081e..06b6092 100644
--- a/homotopy-core/src/diagram.rs
+++ b/homotopy-core/src/diagram.rs
@@ -89,6 +89,11 @@ impl Diagram {
 
     #[must_use]
     pub fn identity(&self) -> DiagramN {
+        #[cfg(debug_assertions)]
+        let (_sig, d) = crate::examples::two_monoid();
+        #[cfg(debug_assertions)]
+        crate::debug::debug_diagram(d.into());
+
         DiagramN::new_unsafe(self.clone(), vec![])
     }
```

and trigger the construction of an identity.


NickHu:
TODO:
* [ ] Make text not blurry and scale correctly
* [ ] Make arrow tips not computed in the stupidest way possible
* [ ] Programatically strip diagram boundaries (currently hardcoded - that's why the tests fail too) - this will require a refactor
* [ ] Maybe generate the geometry separately from how it's generated for normal 3D mode

NickHu:
@doctorn would love a review on this, I'm for sure doing some things incorrectly GL-wise.

jamievicary:
Very interesting idea Nick. How deep will this go into the scaffold?

NickHu:
This currently just uses the geometry you would get if you projected any diagram into 3D, but the intention is that this can be changed on the fly in order to draw any DAG representation of the fully exploded diagram. For instance, we might want to debug a 4D diagram by drawing all its 3D slices in the standard way, but in 3D space offset in the Z-axis (like the 'stack diagrams' we draw sometimes).

homotopy-web/src/app/debug_gl.rs
```diff
@@ -0,0 +1,262 @@
...
+pub struct DebugScene {
+    // rendered by 2D canvas context
+    pub vertex_labels: Vec<(Vec3, String)>,
+    // rendered by GL
+    pub wireframe_components: Vec<VertexArray>,
+}
+
+impl DebugScene {
+    pub fn new(ctx: &GlCtx, diagram: &Diagram) -> Result<Self> {
+        let cubical = match diagram.dimension() {
```
calintat, line 35: `CubicalGeometry::new` isn't really designed for this. What you want is a different constructor that makes a geometry containing all the points/lines but none of areas/volumes. That would be easy to write, and you wouldn't even need to use the `Mesh`, you can just use the exploded graph (and configure that to ignore the boundary).

homotopy-web/src/buffers.rs
```diff
@@ -60,13 +64,13 @@ where
 
     fn push_vert(&mut self, v: B::Vertex, data: VertexData<B>) -> u16 {
         if let Some(&idx) = self.mapping.get(&v) {
-            return idx;
+            idx
```
calintat, line 67: If you use a different geometry which contains all the vertices as points (see my comment above), you wouldn't need to hack anything in buffers.

homotopy-web/src/buffers.rs
```diff
@@ -573,6 +577,104 @@ impl Bufferer for CylinderWireBufferer {
...
+    type Output = DebugArrayData;
+    type State = DebugBuffererState;
+    type Key = ();
+
+    fn new(_geom: &SimplicialGeometry) -> Self {
+        Self
+    }
+
+    fn buffer(ctx: &mut BufferingCtx<Self>) -> Result<()> {
+        for (n, (tri, _)) in ctx.geom.areas.values().copied().enumerate() {
```
calintat, line 620: If you use a different geometry which contains all the points/lines (see my earlier comment about), you wouldn't need this horrible hack.

NickHu: The reason this is here is because in what gets drawn, each edge is pushed together (to make a gap for the node label). This means that there's no longer a one-to-one correspondence between logical points in the geometry and points that get drawn by the renderer, rather each logical point spawns a cloud of rendered points, one for each edge touching it.

calintat: Yeah I get that. What I meant is you're using the triangles to construct the points/lines (which is ugly) and you won't need to do once you do what I said in my comment above.

NickHu: Oh you were talking about the areas part rather than the enumerate part. Yeah, I see that. I think keeping these indices is probably a hack too but I'm not too sure how to fix it.

calintat: Also the correct way to do this is to "inflate" the geometry before you pass it to the bufferer, like Nathan did for cylinders and spheres.

doctorn:
I'll take a look tomorrow!

NickHu:
I fixed the resolution problems with the text by making the 2D canvas which is drawn on top get rescaled. I also added some basic occlusion of text labels:
image
Unfortunately, this occludes over everything in the 3D scene, which is unavoidable with this method of text rendering. I'm not really sure if this occlusion is worthwhile as a result.

Pictured is the associator; it looks a bit yikes but it's a lot nicer interactively rotating around the scene.

Issue #707, opened 2022-08-18
Run-time error when trying to view a very big knot in 3D


H-Yulong:
The application panicked when I was trying to view the Goeritz knot in 3D.
[Goeritz.zip](https://github.com/homotopy-io/homotopy-rs/files/9374305/Goeritz.zip)

calintat:
I'm hoping this will be fixed when we improve layout performance, as I am sure I was able to render your Goeritz knot a while ago when layout was faster.

calintat:
But these knots are great for benchmarking mesh/layout, thanks!

calintat:
Sadly we can no longer import Yulong's knot proofs as the data structures are incompatible. @H-Yulong if you feel like redoing the proofs now with invertible generators, that would be great!

H-Yulong:
Sure I'll see what I can do!

PR #760, opened 2022-08-29
Converted layout to using good_lp

regular-citizen:
The goal is then to switch to an underlying highs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -720,6 +720,7 @@ checksum = "b51d78cbb7b734379eea7f811ddb33b2b13defefa1dab50068d7bc7f781a3056"
 dependencies = [
  "fnv",
  "highs",
+ "minilp",
 ]
 
 [[package]]
@@ -813,13 +814,13 @@ dependencies = [
  "criterion",
  "flate2",
  "getrandom",
+ "good_lp",
  "hashconsing",
  "highway",
  "homotopy-common",
  "insta",
  "itertools",
  "log",
- "minilp",
  "once_cell",
  "petgraph",
  "proptest",
--- a/homotopy-core/Cargo.toml
+++ b/homotopy-core/Cargo.toml
@@ -16,7 +16,6 @@ getrandom = { version = "0.2.7", features = ["js"] }
 hashconsing = "1.5.1"
 highway = "0.8.0"
 log = "0.4.16"
-minilp = "0.2.2"
 once_cell = "1.13.1"
 petgraph = "0.6.2"
 serde = { features = ["derive"], version = "1.0.144" }
@@ -27,6 +26,7 @@ flate2 = "1.0.24"
 serde_json = "1.0.85"
 base64 = "0.13.0"
 arbitrary = { version = "1.1.3", optional = true, features = ["derive"] }
+good_lp = { version = "1.3.2", features = ["minilp"], default-features = false}
 
 [dev-dependencies]
 cargo-husky = { features = ["precommit-hook", "run-cargo-check", "run-cargo-test", "run-cargo-fmt", "run-cargo-clippy"], version = "1.5.0" }
--- a/homotopy-core/src/layout.rs
+++ b/homotopy-core/src/layout.rs
@@ -1,11 +1,14 @@
 use std::ops::{Deref, DerefMut};
 
+use good_lp::{
+    default_solver, variable, Constraint, Expression, ProblemVariables, Solution, SolverModel,
+    Variable,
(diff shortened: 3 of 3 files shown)


calintat:
This is amazing, thanks Chiara! What's stopping us from switching to HiGHS now?

regular-citizen:
In theory "just switching feature flags on good_lp". I'm writing a patch to do that cleanly.

I've given a highs version a try, and I have some good news and some bad news. Good news is that it is way faster than minilp, and in a couple seconds can work out the 4d rendering of Eckmann-Hilton. Bad news is that it gets layout wrong, in the sense that some wires go completely off.

Three things that could be going on with that:
- The ".sorted" lines that I removed were actually necessary (the only places that use that are commutative operations),
- I messed up the weak constraints (I checked and I'm using .geq properly, I think).
- Layout is underspecified as a LP problem, and we were simply used to the "minilp" flavour of solutions.

I'll shortly upload a patch to show what that looks like.

calintat:
Fantastic news! It is definitely the case that layout is underspecified as a problem. In fact there are several open issues about further "desirable" constraints we could impose which would reduce the solution set. I don't think the sorting made any difference, if I recall correctly it was there for debug purposes.

I checked your code and everything looks correct to me. Moreover, this passes the tests so clearly it gives the same solutions as the version on master. I suggest we merge this now and then we figure out how to get it to work with HiGHS.

regular-citizen:
> Fantastic news! It is definitely the case that layout is underspecified as a problem. In fact there are several open issues about further "desirable" constraints we could impose which would reduce the solution set. I don't think the sorting made any difference, if I recall correctly it was there for debug purposes.
> 
> I checked your code and everything looks correct to me. Moreover, this passes the tests so clearly it gives the same solutions as the version on master. I suggest we merge this now and then we figure out how to get it to work with HiGHS.

It passes the tests because this version is still running minilp (under good_lp). The highs version is not "slightly off", it makes EH unworkable as a 2d diagram.

regular-citizen:
CopyQ hYRbjJ

doctorn:
I know nothing about constraint programming but it looks like some of the nodes near the target aren't being constrained to appear above their inputs?

calintat:
> > Fantastic news! It is definitely the case that layout is underspecified as a problem. In fact there are several open issues about further "desirable" constraints we could impose which would reduce the solution set. I don't think the sorting made any difference, if I recall correctly it was there for debug purposes.
> > I checked your code and everything looks correct to me. Moreover, this passes the tests so clearly it gives the same solutions as the version on master. I suggest we merge this now and then we figure out how to get it to work with HiGHS.
> 
> It passes the tests because this version is still running minilp (under good_lp). The highs version is not "slightly off", it makes EH unworkable as a 2d diagram.

Yes that's what I meant. Clearly your PR implements the "same" minilp problem because otherwise it would give different results, so you must have implemented the same constraints I did.

calintat:
> CopyQ hYRbjJ

Woah, that is weird! Do smaller examples work fine?

calintat:
> I know nothing about constraint programming but it looks like some of the nodes near the target aren't being constrained to appear above their inputs?

Layout is implemented inductively so what we're seeing with the "nodes not appearing above their inputs" is because the 1D layout is off.

regular-citizen:
Some tiny ones do:
CopyQ beprbU
CopyQ WysEUb

Others don't:
CopyQ ZtDHIB

regular-citizen:
Smallest divergence is braid_5.
This is HiGHS layout:
CopyQ ULUqaw

This is old layout:
CopyQ TAstoe

I've even managed to get a failure to solve the 3d layout for an identity wire.

calintat:
Great, thanks for the examples Chiara! I think I may have a hunch or two how to fix this.

Could you please push your patch switching to HiGHS? If you want, you could merge this PR and make a new one.

calintat:
Also I'm dying to know how much faster this really is. Do any of the knots in #707 work?

regular-citizen:
> Great, thanks for the examples Chiara! I think I may have a hunch or two how to fix this.
> 
> Could you please push your patch switching to HiGHS? If you want, you could merge this PR and make a new one.

Let's do this.

> Also I'm dying to know how much faster this really is. Do any of the knots in #707 work?

Haven't tried and doing a clean switch to highs is a bit harder than expected due to feature flag woes (that I'm patching up). I also accidentally closed the tab in which I had the highs version, so you're gonna need to wait 5 mins to know.

PR #761, opened 2022-08-29
Enable HiGHS

regular-citizen:
In this PR we will straighten out layout bugs due to highs, fix tests and work out a sensible fallback strategy for binary builds.


regular-citizen:
Calin, I tested #707 knots on the preview. It can do 3d renderings of Goeritz, but 4d crashes with "invalid problem".

For hunters bowknots, it can do 3d and 4d renders effortlessly.

regular-citizen:
Goeritz still crashes with invalid problem.

PR #771, opened 2022-08-31
hover function prototype

H-Yulong:
Prototype the function we want in #11.


calintat:
Thanks Yulong! I pushed a commit replacing the screen coordinate with the generator name.

calintat:
> Great! Shall we merge this?

Yes please!

PR #776, opened 2022-08-31
Remove `util.rs`

doctorn:


calintat:
Thanks Nathan! I actually had a version of this at one point in labelled.

Personally I think it might be cleaner to just use `rev().max_by_key(|g| g.dimension)` everywhere instead the trait.

Also be aware that when you merge this, it will cause a merge conflict with labelled so you'll have to rebase labelled as well.

doctorn:
> Personally I think it might be cleaner to just use `rev().max_by_key(|g| g.dimension)` everywhere instead the trait.

I would tend to agree with you if it wasn't for the need for `rev` which without documentation seems very strange.

> Also be aware that when you merge this, it will cause a merge conflict with labelled so you'll have to rebase labelled as well.

Ah ok! Will do.

calintat:
> I would tend to agree with you if it wasn't for the need for `rev` which without documentation seems very strange.

Ok fair enough!

calintat (approved):

PR #783, opened 2022-09-02
Correct colours for all renderers

### LINKED ISSUE #550: Orientation information in the renderer

thud:
Closes #550 once and for all, implementing correct colours for 3D, 4D and remaining 2D renderers.

Builds on #717.


thud:
I have done my best to implement correct colours for the GL renderer. However, I'm not 100% sure that the renderer works as expected. I would be grateful for some testing if anyone wants to try it.

homotopy-graphics/src/geom.rs
```diff
@@ -98,6 +98,7 @@ pub struct VertData {
     pub position: Vec4,
     pub boundary: [bool; 4],
     pub generator: Generator,
+    pub k: usize,
```
doctorn, line 101: Perhaps you could introduce a typealias or new type for this `usize`? What's the maximum value likely to be (could we get away with a `u32` or `u16`)? `k` is also not a very informative name haha

doctorn (changes requested):
Well done navigating the buffering code! Could you move `buffers.rs` into `src/diagram_gl`?

homotopy-graphics/src/geom.rs
thud: Yeah very true. I'll get something better implemented soon.

homotopy-graphics/src/style.rs
```diff
@@ -51,16 +46,16 @@ impl Color {
     // Get color in lightened form. We wrap colors if needed so that they are sufficiently
     // distinguishable for end users.
     //
-    // Colors are calculated from C and R where:
+    // Colors are distinguished as pairs (C, R) where:
     //      C = max(0, D - N - K)
```
doctorn, line 50: Could you `println` the values of $C$ for some diagram? $K = max(0, D - N)$ so I have a feeling this is always $0$.

thud: In think $K$ is technically not defined a $\max(0, D - N)$. It's actually inferred from the representation of a generator which comes out of the core graphics code.

thud: 2022-09-05-082305_786x419_scrot

In this case, `Cell 2` gives $C = 0$ but `Cell 0` gives $C = 1$.

`hom` attached: [c_values.zip](https://github.com/homotopy-io/homotopy-rs/files/9487626/c_values.zip)

PR #797, opened 2022-09-12
Login UI

H-Yulong:
User authentication service and UI.
You can now login/logout smoothly.


calintat:
This looks quite good, thanks Yulong!

I think we don't need the top button for logging out, it's fine to just have it as a normal button (like you already do).

Also I'm not sure if we should have yet another drawer for account management. I think it's fine to just put this in the "Project" drawer, either above or below the project metadata and import/export buttons. Maybe with a line separating the project and account sections.

jamievicary:
Hi Calin, we decided to do it like this during the standup last week. There is a lot to go in there potentially.

calintat:
Ok that's fine, I just think we're getting to the point where we have too many drawers.

Makefile.toml
```diff
@@ -79,4 +79,4 @@ args = [
 ]
 
 [tasks.serve]
-run_task = { name = ["build-highs", "dist-static-watch", "build-wasm-watch", "devserver"], fork = true, parallel = true }
+run_task = { name = ["dist-static-watch", "build-wasm-watch", "devserver"], fork = true, parallel = true }
```
regular-citizen, line 82: I'm curious about this change. Is the highs build broken?

H-Yulong: Nah, I just disabled it because I'm not using highs and it is taking my VM's disk space. Will add it back before merging.

regular-citizen: Cool, makes sense.

calintat:
@H-Yulong I have a small request: could you please swap the Account and Project buttons? It makes more sense if the ordering is Account > Project > Signature as that's the correct hierarchy.

calintat:
> I think we don't need the top button for logging out, it's fine to just have it as a normal button (like you already do).

And feel free to ignore this, I changed my mind.

Issue #614, opened 2022-07-23
Actions are unnecessarily added to the undo/redo stack


calintat:
For example, if you select a generator from the signature while there is a diagram in the workspace, that does nothing, but it's still added to the history so you can technically undo it. Similar for some other actions.

Generally, if an action doesn't do anything, it shouldn't just return `Ok(())`, it should return an error. That usually means the action shouldn't have been triggered in the first place.

regular-citizen:
This bug seems to only apply to mouse selections, but not keyboard shortcuts.

doctorn:
I think we just need some smarter logic here:

https://github.com/homotopy-io/homotopy-rs/blob/c99330dd7c913f413f83692b0ec26d1f0df7eb6b/homotopy-web/src/model/proof.rs#L175-L181

calintat:
Actually we need to carefully define when actions are valid (see Action::is_valid). If an action is invalid it won't be added to the undo/history stack.

calintat:
For the SelectGenerator action I mentioned, we don't do that so that's the problem.

regular-citizen:
Is this still open? I cannot reproduce it anymore for SelectGenerator.

There is still the "letter-wise actions" for metadata, but that is part of #745.

calintat:
Fixed by a34efaba30a49fbcda4e1a895bc7ec7982465226.

Issue #745, opened 2022-08-25
Metadata UI bugfixes


regular-citizen:
Now that the metadata fields are merged, we should straighten out some remaining bugs with it:
- [] Implement wrap around, at least for the abstract field,
- [] Generate history events on a "edit done" basis, instead of character by character,
- [] Fix editing behaviour when cursor is at "inner character", i.e. if I start editing at the first letter, it shouldn't jump at the last.

H-Yulong:
Also need to get rid of the horizontal scroll bar at the bottom of the signature drawer.

- [ ] Fix `overflow-x` for signature drawer

calintat:
Also I like how for generator names, we underline the text when it's editable. We should do the same for the metadata. 

@thud could you help with that since you did it for generator names.

thud:
Yep, can do, think that's just a case of styles.

Also I should mention that `item.rs` implements a method of avoiding generation of history events for rename events (might be helpful for anyone attempting that). I'm also happy to give that a crack if needed.
https://github.com/homotopy-io/homotopy-rs/blob/878bc9b18f2529b160518d0f2b638c02c79c26ef/homotopy-web/src/app/signature/item.rs#L224-L233

regular-citizen:
Thanks for sharing this snippet Jasper!

thud:
I'm also going to add the following issue to the list (assuming others feel the same).

- [ ] The inputs have a default value which has to first be deleted before you can change the input to something useful. Could we replace this with placeholder text?

jamievicary:
That sounds very good. to me Jasper.

thud:
> Also need to get rid of the horizontal scroll bar at the bottom of the signature drawer.
> 
>     * [ ]  Fix `overflow-x` for signature drawer

@H-Yulong, can you check to see if this is fixed in #678?

calintat:
One more thing and this is more of nitpick. It looks like the abstract isn't horizontally aligned with the title and author.

H-Yulong:
> 

Yes it is.

thud:
This should be closed by #808.

PR #808, opened 2022-09-13
Add to metadata UI

### LINKED ISSUE #807: Title in top-left corner
We should display the project title in the top-left corner of the workspace in a light grey.

This way the title will be visible even if the Project panel is closed.

thud:
Adds fixes to metadata UI (including text wrapping). Adds large project title to top left corner of workspace.

Closes #807, #745.


thud:
I don't think there is a good way of making the metadata UI inputs resize themselves (for multiline text). In my opinion, textarea is the best solution but happy for feedback. Otherwise, this should be ready for merging.

calintat:
Thanks Jasper, this looks really good!

One comment: I just noticed that if you click on the title, don't change anything, and then click somewhere else, it still adds an EditMetadata action to the history. Perhaps we should only add the action to the history **if** the new metadata is not equal to the old metadata.

thud:
Oh yeah, good catch. I'll fix this now.

calintat:
One more thing (this is very nitpicky sorry): if the title is very long, maybe we should cut it off with an ellipsis so it doesn't go under the slice controls.

thud:
Alright @calintat, is this what you meant with the ellipsis?

regular-citizen:
Thanks Jasper, looks to me that this will also close #614.

calintat:
> Thanks Jasper, looks to me that this will also close #614.

#614 is something else, it's not about the metadata

thud:
@calintat is correct I think.

calintat:
> Alright @calintat, is this what you meant with the ellipsis?

Yes, but now the slice control moves to the left if there is no title.

thud:
I believe this is now fixed.

calintat:
The preview hasn't been updated, but I believe you.

Feel free to merge this if you want.

PR #844, opened 2022-09-23
Allow Tex input where applicable

thud:


thud:
For whatever reason, the `katex` crate doesn't seem to pass `cargo check` with the `wasm-js` feature enabled. I'm not currently sure why.

thud:
- [ ] Enable tex input for generator names.

calintat:
Great job, thanks Jasper! I have a few comments.

The generator name looks a bit off. It's not really vertically centred and it has some extra space to the left which I don't like:

Screenshot 2022-09-24 at 12 22 05 Screenshot 2022-09-24 at 12 22 15

Also, in the project metadata, the font for "Title" is way too big, can we go back to what it was before. Also the metadata gets clipped if it's longer than one line (when not editable) which is not good.

calintat:
> [ ] Enable tex input for generator names.

What do you mean by this?

calintat:
We're missing the TeX rendering in two places:

- [ ] The project title in the top-left corner of the workspace.
- [ ] The tooltips when you hover over a generator in the diagram.

regular-citizen:
> For whatever reason, the `katex` crate doesn't seem to pass `cargo check` with the `wasm-js` feature enabled. I'm not currently sure why.

They have a dumb Cargo.toml. It will only add wasm-bindgen/js-sys as a depencency if you're targetting wasm,
even though both dependencies are designed to check/compile even outside the browser.

``cargo check --target wasm32-unknown-unknown`` works fine.

thud:
Yeah quite right. There are a few things to fix still.

Notably, the title no longer uses ellipsis properly for long titles / thin workspaces.
Also, I'm not sure how possible it is to render non-text elements in the `<title>` element as it's supposed to be used for accessibility reasons. We might have to refactor this into a component instead (a floating `<div>` or whatever) if we want rendering for it.

calintat:
One more thing: can we change the font size of the rendered LaTeX? It looks slightly bigger than the rest of the text (compare 0 vs $0$).

I kinda like how the workspace title looks like without the ellipsis tbh, but it forces the view control to be the same height.

Screenshot 2022-09-26 at 13 44 26

thud:
Just realized, I also need to implement proper error handling for the latex rendering. Shouldn't be too difficult...

calintat:
Thanks Jasper!

The project metadata is still not rendered properly if it is too long. See this workspace: [homotopy_io_export.hom.zip](https://github.com/homotopy-io/homotopy-rs/files/9656042/homotopy_io_export.hom.zip)

calintat:
I just rebased the branch on top of my commit e21752e50245f5994c0e7b828c087cd9c792ac81 and I think I also fixed the problems with undo while doing that.

Issue #818, opened 2022-09-14
Import/export actions in debug panel


calintat:
We should have buttons for importing and exporting actions in the debug drawer (which is only visible in debug builds).

This could replace the "Dump actions" button which simply dumps the actions to the console. Also I think we should remove the option to import actions from the main "Import" button in the project drawer as we don't want users to accidentally import .txt files.

PR #846, opened 2022-09-26
Action improvements

regular-citizen:
This PR will address a bunch of open issues regarding actions.


regular-citizen:
Right now, the changes introduced have two problems, compared to the earlier solution.
- It breaks every existing action trace, which need to be changed as in the first commit on this PR. It's a quick thing, it suffices to change ``$0`` with ``[false,$0]``, where ``$0`` were the contents of the earlier action trace.
- As a consequence of ImportProof being a proof::Action, importing an action dump now results in a single action being created, so we can no longer step through an imported trace.

(ii) is really annoying. I think a way to solve it would be to change the return type of update to Result<Vec<Action>,ModelError>, so that we can Ok(vec![]) with the actions actually performed and add potentially multiple actions to history. This should also have the side effect of effectively rendering the current "is_valid" check useless, as we can just try actions and if they fail they don't get added to history.
However this strategy will 100% cause problems when rebasing labelled, so I'm not yet rushing to implement it. @calintat do you have thoughts on this?

calintat:
Nice, thanks Chiara!

Why is (ii) happening? Is it because the functionality to import actions is part of the `ImportProof` action so it automatically adds an `ImportProof` action to the history. Because I think we should remove the functionality to import actions from `ImportProof` (see #818) and have that be a separate thing (not a proof action).

calintat:
I'm not a big fan of the separators in the debug panel. Also it would make sense to put "Export actions" and "Import actions" on the same row.

regular-citizen:
> I'm not a big fan of the separators in the debug panel.

How would you like them to look? I'm using those because they're essentially pre-made, and otherwise it looks quite bad.

> Also it would make sense to put "Export actions" and "Import actions" on the same row.

Got it. Should "dump diagram" and "dump signature" be in the same row as well?

PR #854, opened 2022-09-29
Catch firebase errors

### LINKED ISSUE #835: Beta broken, likely due to firebase-compat CORS failure
Opening this in case it wasn't picked up.

Beta is broken on my machine (Firefox/Linux), in the sense that yew app doesn't even start.
For some reason an error in:
https://github.com/homotopy-io/homotopy-rs/blob/master/homotopy-web/static/index.html#L43
is stopping the yew app's main from running.

Console dump:
- Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://www.gstatic.com/firebasejs/9.7.0/firebase-performance-compat.js. (Reason: CORS request did not succeed). Status code: (null)
- Loading failed for the module with source “https://www.gstatic.com/firebasejs/9.7.0/firebase-app-compat.js”.
- Loading failed for the module with source “https://www.gstatic.com/firebasejs/9.7.0/firebase-analytics-compat.js”
- Module source URI is not allowed in this document: “https://www.gstatic.com/firebasejs/9.7.0/firebase-performance-compat.js”.
- Loading failed for the module with source “https://www.gstatic.com/firebasejs/9.7.0/firebase-auth-compat.js”.
- TypeError: error loading dynamically imported module

regular-citizen:
Fixed #835.

### DIFF
--- a/homotopy-web/static/firebase.js
+++ b/homotopy-web/static/firebase.js
@@ -1,21 +1,5 @@
 // This script initializes firebase.
-
-console.log("Loading firebase...");
-
-// Compat modules for firebase ui
-import "https://www.gstatic.com/firebasejs/9.7.0/firebase-app-compat.js";
-import "https://www.gstatic.com/firebasejs/9.7.0/firebase-analytics-compat.js";
-import "https://www.gstatic.com/firebasejs/9.7.0/firebase-performance-compat.js";
-import "https://www.gstatic.com/firebasejs/9.7.0/firebase-auth-compat.js";
-
-// Import the functions you need from the SDKs you need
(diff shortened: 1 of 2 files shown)


regular-citizen:
I think this should fix the problem, but I'm not too confident with firebase API, so @H-Yulong I'd really appreciate your feedback on this. If it looks good to you we can merge it immediately.

H-Yulong:
Sure! Let me have a look at this.

H-Yulong:
It works fine! I think we can merge this.
(I deleted the error messages in `index.html` since they're useless now)

PR #855, opened 2022-09-29
Enable compression.

### LINKED ISSUE #850: Compress .hom files
As discussed today, it might be good to compress the .hom files before uploading them to Firebase, as it saves our cloud storage space and it is not hard to implement.

I did some experiments on my knot files.

```
> Culprit.hom    672.0KB -> 181.9KB
> Goeritz.hom    483.4KB -> 135.2KB
> H Knot.hom    376.4KB -> 111.1KB
> Monster.hom    373.0KB -> 111.4KB
> Hunters Howknot.hom    370.8KB -> 113.6KB
> J Knot.hom    361.4KB -> 109.1KB
> Knot Theory Template.hom    280.3KB -> 85.5KB
> Trivial Knot.hom    73.7KB   -> 19.2KB
> Empty hom file    156 bytes -> 144 bytes
```

It turns out that zlib does a pretty good job of reducing file size and we should definitely compress the .hom files before uploading.

Also, is it a good idea that we always compress/decompress when we're exporting/importing?

regular-citizen:
Fixes #850.

Deseralization code includes a fallback, so we're still able to import old .homs without version bumps.
To compress .homs, simply run them through homotopy-cli, as ``homotopy-cli -i input.hom -o output.hom``.


NickHu:
I switched it from brotli to zstd, based on our discussion at #850. We should think up a principled way to generate a dictionary perhaps, rather than just running `zstd --train` on our examples directory.

calintat:
Perhaps we could train the dictionary on the workspaces stored on Firebase right now (i.e. the ones from the JS version) to get a good sample of "likely" workspaces.

NickHu:
The issue with that is that it will then train itself on the old format, which we don't use anymore.

calintat:
I meant after migration.

regular-citizen:
I agree that we should train on the full migrated dataset, but I would merge this PR with a temp dictionary trained on examples, to be replaced after we merge labelled, which is gonna change a lot of data structures.

regular-citizen:
I'd be happy to write the patch to include a temporary dictionary, it takes me 5 mins.

calintat:
@NickHu so does this automatically use everything in `examples/` as the dictionary? Because I think that's quite risky: if we ever add another example or change one of the existing examples (which we will have to do e.g. when we merge labelled) that will break the decompression.

regular-citizen:
> @NickHu so does this automatically use everything in `examples/` as the dictionary? Because I think that's quite risky: if we ever add another example or change one of the existing examples (which we will have to do e.g. when we merge labelled) that will break the decompression.

I was under the mistaken impression that decompression would not need a dictionary.
I believe then the best course of action would be to merge labelled, migrate the JS database, and then compute the dictionary on that. We will then merge this PR with the changes implementing the dictionary.

calintat:
Actually I don't think using the JS projects for the dictionary is a good idea anymore. Because when they're migrated, everything is treated as oriented and not framed (i.e. no labels). So that's not very representative of the workspaces people are likely to build in practice.

calintat:
Honestly I'm not sure what a good choice of a dictionary would be, maybe it's best to not use a dictionary or use a really simple one.

regular-citizen:
> Honestly I'm not sure what a good choice of a dictionary would be, maybe it's best to not use a dictionary or use a really simple one.

We could just merge like this, without dictionaries. Your point about orientations makes a lot of sense.

Issue #523, opened 2022-06-20
Combine signatures


calintat:
In the "Signature" panel, we should be able to insert the signature of a different workspace to our current signature. This is different from the import functionality which replaces the current project so it doesn't let you combine signatures.

This way, we could easily add structures such as monoids to the signature.

calintat:
Maybe place the inserted signature in a folder.

regular-citizen:
Do we have some UI in mind for how to do this without creating a "checkbox hell"?
I really like the idea of "import workspace inside a new folder", but I still can't picture a non-clunky UI.

jamievicary:
Drag the .hom file onto the directory icon?

regular-citizen:
I like this idea. Though if we do it for folders we should also support it for old-school "replace workspace" imports.

mlharaujo:
It would be nice if you could also choose how to attach the new cells. For example, you might want to add a monoid structure to an already existing cell (or even to a diagram).

jamievicary:
The way to do this would be by first importing the external signature by disjoint union, then identifying generators with the same boundary.

doctorn:
This is a restricted push-out of computads right?

jamievicary:
Hi Nathan yes definitely that would be the expectation.

mlharaujo:
What do you mean by "restricted"?

doctorn:
Restricted in the sense that extending a signature like this will only provide you with a subclass of all push-outs of computads (my hunch is that not all of them will be constructible without assuming the existence of inverses and postulating coherence data — something we can only do on the labelled branch).

jamievicary:
This is something we do not yet give a full mathematical treatment of, however it is clear that identifying two variables with identical boundaries is a very simple sort of pushout, which can be computed more straightforwardly than other richer sorts of pushouts one might consider.

PR #857, opened 2022-10-02
Integrated FileSaver.js

regular-citizen:
SaveAs doesn't really do saving as on Firefox/Linux and Chromium/Linux.

### DIFF
--- a/homotopy-web/src/model.rs
+++ b/homotopy-web/src/model.rs
@@ -5,7 +5,7 @@ pub use homotopy_model::{history, migration, proof, serialize};
 use js_sys::JsString;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
-use wasm_bindgen::{prelude::*, JsCast};
+use wasm_bindgen::prelude::*;
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub enum Action {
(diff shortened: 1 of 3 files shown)


calintat:
This doesn't work on my Mac on any of Safari, Firefox, or Chrome.

NickHu:
This library hasn't been updated in two years. There's a new fs API, and basically the call we want is https://developer.mozilla.org/en-US/docs/Web/API/Window/showSaveFilePicker. Unfortunately, it's experimental and isn't available on all browsers. Fortunately, it can be polyfilled quite easily: https://github.com/jimmywarting/native-file-system-adapter/

regular-citizen:
> This doesn't work on my Mac on any of Safari, Firefox, or Chrome.

Yeah it was mostly a demo of "saveAs" actually just being a polyfiller for "downloadWithThisGivenName".

regular-citizen:
> Fortunately, it can be polyfilled quite easily: https://github.com/jimmywarting/native-file-system-adapter/

I didn't know about this, this is very cool. It seems that it should also be able to handle drag and drops, which we could use for imports.

calintat:
> > Fortunately, it can be polyfilled quite easily: https://github.com/jimmywarting/native-file-system-adapter/
> 
> I didn't know about this, this is very cool. It seems that it should also be able to handle drag and drops, which we could use for imports.

Oh that would be really cool. We could also let people drag and drop a .hom file into a signature (#523).

regular-citizen:
I'll close this PR and have a crack at integrating that in the coming days.

PR #860, opened 2022-10-02
Save As and Drag Drop functionalities

regular-citizen:


calintat:
Is this meant to work? Because it doesn't for me :(

regular-citizen:
> Is this meant to work? Because it doesn't for me :(

It doesn't work for me either. I'm not sure if there is any way of getting the "save as" functionality to work with this.
The current implementation tries to trigger a "save as" download only after a user interaction, which seems to be required by browsers to actually display the prompt, but it still doesn't work.

NickHu:
It works for me, but only locally and not on the Firebase preview. In the preview I get this error in the console:

```
Uncaught ReferenceError: gen_download is not defined
    __wbg_gendownload_58732fa669d97ce0 https://homotopy-rs--homotopy-io-saveas-dragdrop-9q5nra2n.web.app/homotopy_web.js:410
    __wbg_adapter_47 https://homotopy-rs--homotopy-io-saveas-dragdrop-9q5nra2n.web.app/homotopy_web.js:273
    real https://homotopy-rs--homotopy-io-saveas-dragdrop-9q5nra2n.web.app/homotopy_web.js:255
homotopy_web.js:410:21
```

regular-citizen:
> It works for me, but only locally (tested Firefox Linux) and not on the Firebase preview.

This is very weird. Can you check if the browser is loading at all "native_fs_adapter.js"?

NickHu:
Ah, disabling uBlock made it work.

regular-citizen:
> Ah, disabling uBlock made it work.

Yeah we should just make a copy of the library and serve it from our servers.
By the way, by "it works for me", do you mean you actually get shown a file picker?

NickHu:
Yes, the file picker works for me

calintat:
Interesting it doesn't work for me.

calintat:
I didn't check but it's not that it panics or anything like that, it just downloads a file without opening the save-as dialog.

Issue #553, opened 2022-06-27
Arrows on wires


calintat:
This will be necessary to display the orientation of a wire (#550).

It would also be nice to have as an option in the non-invertible version.

thud:
Ah yes of course! Will do now.

PR #864, opened 2022-10-07
New show orientation

thud:
Add arrows / asymmetric boxes to invertible generators.


thud:
https://github.com/homotopy-io/homotopy-rs/issues/553#issuecomment-1269630734

This is a pretty messy commit which represents a few days of random prototyping with arrows and boxes. As I recall, the arrows just point randomly at the moment and the same, I think, applies for the boxes.

Hopefully it is of some use to someone.

Another note, I'm not sure if this plays well with the path simplification algorithm (used by `tikz` renderer etc.).

Issue #911, opened 2022-11-21
Tool crashed and "Donwload action log" button doesn't work


mlharaujo:
I opened the console and copied the error messages to this notepad. Unfortunately I don't have the proof state I was working on, since I wasn't expecting a crash.

[console.txt](https://github.com/homotopy-io/homotopy-rs/files/10057377/console.txt)

![image](https://user-images.githubusercontent.com/36002049/203081105-4daaeda0-ec81-4109-984c-60e8d84f4823.png)

regular-citizen:
Oof this is due to a out of memory error ("Uncaught InternalError: allocation size overflow" showing up in the logs) on the JS side when dumping out the log file. I'm amazed that action dumps are that big, but it makes sense that it could happen.

The actual panic was cause by Cone::new in homotopy-core/src/rewrite.rs. In particular the first assertion of the method seems to be exploding:
https://github.com/homotopy-io/homotopy-rs/blob/98f4ef7d9dd4b4e06a406a686d3ee8ecad8f1d0c/homotopy-core/src/rewrite.rs#L906

regular-citizen:
Oh I think the culprit is that now imports get included in the action dumps (as part of #831 fix). That means that if you reimport the same meg-sized .hom file multiple times, an action dump will tend to error out of memory. That should be a quick fix though.

The real fix would be to try not to store action dumps in RAM but in local storage.

mlharaujo:
So this explains the problem with the action log, but the not the crash itself, right?

regular-citizen:
Yes, that is the case.

mlharaujo:
A similar thing has happened again, I'm not sure if it's the same problem. When I try to do an expansion in the diagram below, to resolve the half braid into a full braid, the tool crashes. I get the spinning wheel and no option to download action logs. 
![image](https://user-images.githubusercontent.com/36002049/204244184-1b657499-31e3-4f10-a96c-f08b32c52e9a.png)

I'm attaching the .hom file and the console output.

[coherence_laws_2.zip](https://github.com/homotopy-io/homotopy-rs/files/10102341/coherence_laws_2.zip)
[console.txt](https://github.com/homotopy-io/homotopy-rs/files/10102345/console.txt)

mlharaujo:
Same thing happened again, but this time there is a very short console output. I had clicked the top right dot and dragged down in the diagram below.
[...]

PR #918, opened 2022-11-26
Refactored panic handling logic to avoid OOMs

regular-citizen:
This PR refactors significantly the panic handling logic, and the associated code to store action dumps.
It aims to significantly reduce the memory footprint of dumps to avoid problems associated with longrunning usage of the tool, such as in #911.

### DIFF
--- a/homotopy-model/src/history.rs
+++ b/homotopy-model/src/history.rs
@@ -173,6 +173,22 @@ impl History {
         actions
     }
 
+    pub fn get_last_import_segment(&self) -> Vec<super::proof::Action> {
+        let mut actions = Vec::new();
+        for a in self
+            .snapshots
+            .ancestors_of(self.current)
+            .filter_map(|n| self.snapshots.with(n, |s| s.action.clone()).flatten())
+        {
+            actions.push(a.clone());
+            if matches!(a, super::proof::Action::ImportProof(_)) {
+                break;
+            }
+        }
+        actions.reverse();
+        actions
+    }
+
     pub fn last_action(&self) -> Option<super::proof::Action> {
         self.snapshots
             .with(self.current, |s| s.action.clone())
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1044,6 +1044,7 @@ dependencies = [
  "yew",
  "yew-agent",
  "yew-macro",
+ "zip",
 ]
 
 [[package]]
@@ -2392,4 +2393,16 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn",
+]
+
+[[package]]
+name = "zip"
+version = "0.6.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "537ce7411d25e54e8ae21a7ce0b15840e7bfcff15b51d697ec3266cc76bdf080"
+dependencies = [
+ "byteorder",
+ "crc32fast",
+ "crossbeam-utils",
+ "flate2",
 ]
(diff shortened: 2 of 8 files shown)


homotopy-model/src/history.rs
```diff
@@ -173,6 +173,22 @@ impl History {
...
 
+    pub fn get_last_import_segment(&self) -> Vec<super::proof::Action> {
+        let mut actions = Vec::new();
+        for a in self
+            .snapshots
+            .ancestors_of(self.current)
+            .filter_map(|n| self.snapshots.with(n, |s| s.action.clone()).flatten())
+        {
+            actions.push(a.clone());
+            if matches!(a, super::proof::Action::ImportProof(_)) {
```
calintat, line 184: You can use `Iterator::take_while` to make this neater.

calintat:
Thanks Chiara, this looks much better!

homotopy-model/src/history.rs
regular-citizen: I'm not sure I can. I need to include the last ImportProof. I don't think there is a neat way of expressing it.

regular-citizen:
Two more things I wanted to do in this PR:
- Figure out how to actually make the crash banner play nice with the loading spinner, but I need to talk to Nick regarding what hacks he used to do that, because it seems to be overridding my panic handler.
- Potentially write a better serializer for Height. Looking at the hex dumps of .hom files, it looks like most of the data is just strings "Regular/Singular + number in binary" over and over.

calintat:
> Two more things I wanted to do in this PR:
> 
> * Figure out how to actually make the crash banner play nice with the loading spinner, but I need to talk to Nick regarding what hacks he used to do that, because it seems to be overridding my panic handler.
> * Potentially write a better serializer for Height. Looking at the hex dumps of .hom files, it looks like most of the data is just strings "Regular/Singular + number in binary" over and over.

We cannot change the serializer for data structures in homotopy_core::common. Those are fixed forever because diagrams/rewrites depend on the things there. And I think it's best to rely on the default derives.

calintat:
The loading spinner isn't really a hack. It's just a css animation that's rendered whenever `App::loading` is true. The problem is if an action crashes, it never has a chance to set `loading` to false. I would just make sure your panic handler dialog is rendered above the spinner.

regular-citizen:
> We cannot change the serializer for data structures in homotopy_core::common. Those are fixed forever because diagrams/rewrites depend on the things there. And I think it's best to rely on the default derives.

Yes that was a concern. The way I was thinking of going about this was to still include fallback serializers, but it's annoying to have that code around.

> The loading spinner isn't really a hack. It's just a css animation that's rendered whenever `App::loading` is true. The problem is if an action crashes, it never has a chance to set `loading` to false. I would just make sure your panic handler dialog is rendered above the spinner.

Yes, and I can deal with that on the JS side by just removing it from the DOM if a panic occurs. However, I believe there is something happening with the futures used that is preventing me from installing a panic handler on the correct thread. I cannot do it through yew, and I cannot do it through std::panic, which is quite weird.

regular-citizen:
This is good to merge as far as I am concerned. I believe we should still talk about shrinking homs, but we'll do it in a separate occasion.

calintat:
Can you fix the issue I mentioned in my comment above before merging?

regular-citizen:
> Can you fix the issue I mentioned in my comment above before merging?

The take_while one?

calintat:
No the other one. The panic handler history isn't updated correctly because you push actions that may throw an error (e.g. invalid contraction) or that return false (e.g. take identity on empty workspace).

regular-citizen:
> No the other one. The panic handler history isn't updated correctly because you push actions that may throw an error (e.g. invalid contraction) or that return false (e.g. take identity on empty workspace).

Oh weird, I didn't see this one. You're right, I'll handle this.

regular-citizen:
@calintat think this should handle it.

PR #933, opened 2022-12-05
Experiment with hashconsing for labels

calintat:


calintat:
@regular-citizen I haven't noticed any performance improvements from this, but maybe you could check that with your profiling tool. It is backwards compatible with existing examples so if there are any gains at all we should merge it.

calintat:
I also have a second commit which experiments with storing the labels in a map in `Store`: 
https://github.com/homotopy-io/homotopy-rs/blob/d236a403ea9e7a13e113dfb029a954b9eb5e8551/homotopy-core/src/serialize.rs#L28

I haven't pushed it yet as that one breaks backwards compatibility. Weirdly it didn't seem to reduce the size of hom files at all; actually on some examples it had the opposite effect. Though I only tested it on small examples (KB-sized).

regular-citizen (approved):

PR #934, opened 2022-12-05
Experiment with reducing file size by not replicating labels

calintat:


calintat:
I am closing this as it did not have the effect we expected. We can revisit at a later time.

PR #942, opened 2022-12-06
Unsafe version of homotopy.io

calintat:


calintat:
Here is a version of the tool with the wellformedness checks and some typechecking turned off. Note that all of these are assertions that should in principle never fail. This version should be faster than master, but with the risk of bugs going unnoticed.

Issue #930, opened 2022-12-02
Handle actions that take too long


calintat:
As seen in #927, it is possible to get into a situation where an action is running for too long and there is no way to abort that without losing the proof state. I can think of two solutions:
- add a "cancel" button to the spinner which lets you cancel the action without panicking,
- add an explicit timeout to actions which does panic but at least you can recover the proof state using the panic handler.

This will hopefully be less relevant once we make collapse faster, but it would still be good to have.

Any thoughts?

mlharaujo:
This would be great. It's very frustrating to have to wait a lot of time just because I accidentally clicked something.

regular-citizen:
At the technical level, I was thinking that one possible approach to do this without killing the entire prover would be to store a mutable global ```bail_expensive_computations_asap``` flag, and then define some kind of yield function with signature
```fn yield() -> Result<(), ModelError>```, which just checks the flag,

Then everywhere in the codebase we could just do ```yield()?;``` and propagate the failure upwards. This is inspired by what actually co-routine runtimes do for expensive computations.

regular-citizen:
Discussed this with @calintat with outcome that we would need to integrate seriously with the tokio runtime, and add a bunch of calls to yield_now in core, plus support for aborts.

NickHu:
I think tokio is pretty hard to integrate with WASM, at least last time I checked... good luck

PR #944, opened 2022-12-07
Timeout experiment

regular-citizen:
The basic code implementing the idea proposed in #930 is simply realised. However, this still doesn't work as it seems hard to even tell the WASM code that the "quit" event has triggered, as event processing is blocked until the thing we want to kill finishes.


regular-citizen:
Experiment failed, this code requires interacting with a co-routine/futures runtime such as tokio. Closing for now.

Issue #642, opened 2022-07-29
choosing reading direction for morphisms


vinnylarouge:
it would be nice to be able to toggle bottom2top and left2right as options, especially since it seems that string-diagram reading direction is a matter of taste between different fields.

vinnylarouge:
(i might just be a pervert, but recently I've been depicting string diagrams from left to right like they do in estonia)

regular-citizen:
#953 gives the option to export to tikz with the left-to-right direction. In theory, there is no reason not to have a global flag in the settings that allows to toggle the reading direction.

In practice, doing this everywhere is a bit of work/ For example, the slice control on the right would make no sense in left-to-right mode.

PR #953, opened 2022-12-18
Left-to-right tikz exports

regular-citizen:
Add the option to export tikz diagrams with left-to-right reading direction.


calintat:
Thanks Chiara! 

As discussed privately, we will eventually have a global setting for the reading direction (which we already have an issue for in #642) but that requires more work on the UI for things like the slice controls and the drag handler.

PR #966, opened 2023-01-05
Adding a tutorial

regular-citizen:
This is an attempt to write an introductory tutorial covering the basics of how to use the tool. It describes how to prove that every equivalence may be promoted to an adjoint equivalence, which is a fact that every category theorist should be familiar with, and touches on most aspects of the tool (including inverses).

Feedback is very welcome, the actual file is [here](https://github.com/homotopy-io/homotopy-rs/blob/tutorial/TUTORIAL.md).


calintat:
I always assumed we will hide the Debug panel in the public release, so maybe we shouldn't reference it in the tutorial, but I don't know.

TUTORIAL.md
```diff
@@ -0,0 +1,256 @@
...
+
+![Taking the identity on C as source](./doc/img/tutorial4.png)
+
+For the target, we select $F$ and then paste $G$ on top of it by clicking towards the upper edge. Since there is only one generator that $D$ as a source, namely $G$, it is automatically selected by the tool.
+
+![Constructing the composite of F and G](./doc/img/tutorial5.png)
+
+Now, this composite diagram we just constructed has the same boundary as the identity on $C$, which we selected as a source (they both have source and target $C$). This is a necessary condition for the tool to allow us construct a new generator in the signature called *globularity*.
+
+For example, if we didn't compose $G$ on top of $F$, the globularity check would have failed (as the target of $F$ is $D$ and not $C$) and `Target` button would have been hidden by the tool.
```
calintat, line 62: The `Target` button won't actually be hidden. I changed that a while ago so that Source/Target are visible even if the globularity check fails (1 so we avoid running the check unnecessarily, and 2 because it's more intuitive if you get an error message explaining why the globularity check failed).

TUTORIAL.md
```diff
@@ -0,0 +1,256 @@
...
+## Constructing the Adjoint
+
+Our theorem states that we may construct new 2-cells in this signature which satisfy the triangular equations, also called the snake equations, for an adjuction between $F$ and $G$.
+In particular, we may do so without changing one of the natural transformations.
+
+Let's choose to keep $\alpha$ unvaried and to construct some other $\beta'$ such that $\alpha$ and $\beta'$ satisfy the snake equations. We claim that we may take $\beta'$ to be the composite $\beta \circ \alpha^{-1} \circ \beta^{-1}$.
+
+Now, although we could prove the theorem by inserting this $\beta'$ at all the right places in the proof, keeping track of this data is somewhat non-trivial. For good ergonomics, it is much more convenient to save this definition of $\beta'$ as its own element of the signature.
+
+Since we want something that will eventually replace $\beta$, it should at least have the same boundary as $\beta$. So let's select $\beta$ from the signature, to mark it as the current generator.
```
calintat, line 89: I'd replace "current generator" with "current workspace diagram".

TUTORIAL.md
```diff
@@ -0,0 +1,256 @@
...
+- A cell called 'Theorem', which contains the newly-defined $\beta'$,
+- A cell called 'Proof', which contains the path taking the generator $\beta'$ to the diagram we previously constructed.
+
+Hence the 'Proof' cell is really our witness that to what the definition of $\beta'$ is. Notice that is a 3-dimensional object, as indicated by the number on the left of its name. This is expected, as it should be a path between 2-dimensional things.
+
+Let us rename the 'Theorem' generator to $\beta'$, and the 'Proof' generator to '$\beta'$ Definition' for clarity. The signature should now look like this:
+
+![The signature with our new counit](./doc/img/tutorial11.png)
+
+Notice in particular that $\beta'$ looks exactly like $\beta$, and that it also invertible, as we would expect. We did not have to flag it as such, as they tool already checked that it was constructed by composing only invertible generator.
```
calintat, line 113: invertible generator**s**

calintat (approved):
@regular-citizen Looks good, just a few small comments!

regular-citizen:
> @regular-citizen Looks good, just a few small comments!

Thanks for the feedback. I have implemented it.

Issue #512, opened 2022-06-17
Empty workspace information


calintat:
Many apps like text editors show some information when there is nothing open. We should do the same if the workspace is empty. We could show a message like "Click on a generator to select it. To create a new 0-dimensional generator, press the + button".

This would be hugely useful if a user opens homotopy.io for the first time and have no idea where to start.

thud:
This could also include default keybinding information to help new users. (note from standup 21/07/22)

calintat:
Yes! I am thinking something similar to VS code:
<img width="2672" alt="Screenshot 2022-07-21 at 15 09 42" src="https://user-images.githubusercontent.com/16837293/180234576-9347d855-4d56-4932-8531-03ece96e82d9.png">

jamievicary:
Upvote!

PR #973, opened 2023-01-13
Empty workspace message (#512)

regular-citizen:
Feedback welcome


jamievicary:
Thanks Chiara this is great! Could we make it grey please rather than black? Then I will be happy.

regular-citizen:
> Thanks Chiara this is great! Could we make it grey please rather than black? Then I will be happy.

I assumed this referred to the logo, but let me know if you also want the text lighter.

I have a separate patch with a safety mechanism which disables the "?" keyboard shortcut when the signature contains "interesting" stuff. This is to avoid users being teleported away to the tutorial in the middle of a complicated proof. Right now, it seems that the "unsaved work" mechanism is working effectively, but we may want to be extra careful.

Let me know also on this front, otherwise we are good to merge as far as I am concerned.

calintat:
This looks fantastic, thanks Chiara!

I just pushed a small commit reordering the shortcuts, I think we should merge this!

calintat:
> > Thanks Chiara this is great! Could we make it grey please rather than black? Then I will be happy.
> 
> I assumed this referred to the logo, but let me know if you also want the text lighter.
> 
> I have a separate patch with a safety mechanism which disables the "?" keyboard shortcut when the signature contains "interesting" stuff. This is to avoid users being teleported away to the tutorial in the middle of a complicated proof. Right now, it seems that the "unsaved work" mechanism is working effectively, but we may want to be extra careful.
> 
> Let me know also on this front, otherwise we are good to merge as far as I am concerned.

I wouldn't worry too much about users accidentally pressing '?'.

jamievicary:
Thanks Chiara! I was imagining everything being gray including the text.
This makes it a bit less intrusive for expert users.

regular-citizen:
> I was imagining everything being gray including the text.

Done, let me know if this works, otherwise I can make it even lighter.

calintat:
Can we also make the shortcuts grey to match the text?

PR #986, opened 2023-01-23
Enable weak units

calintat:
@jamievicary The preview is live.


jamievicary:
Wonderful thank you!

calintat:
@jamievicary Actually I just realised we could make this into a user setting. Now if you go to beta, there is a checkbox in settings for enabling weak units. I'd recommend using that instead of the preview (just be careful that the settings get reset whenever you refresh).

jamievicary:
Brilliant thank you.

PR #1013, opened 2023-02-10
New devserver

regular-citizen:
It seems that devserver has been deprecated. This is an ahead of time patch that should fix potential CI problems.

Let me know if you have other devserver preferences. **Default port on sfz is 5000**. I did not test the Makefile.

### DIFF
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -65,8 +65,8 @@ mkdir -p dist
 command = "sfz"
 args = [
   "--render-index"
-  "--coi"
-, "dist"
+,  "--coi"
+, "dist/"
 ]
 
 # for testing Firebase components locally
--- a/Makefile.toml
+++ b/Makefile.toml
@@ -62,14 +62,11 @@ mkdir -p dist
 '''
 
 [tasks.devserver]
-command = "devserver"
+command = "sfz"
 args = [
-  "--path"
+  "--render-index"
+  "--coi"
 , "dist"
-, "--header"
-, "Cross-Origin-Opener-Policy=same-origin"
-, "--header"
-, "Cross-Origin-Embedder-Policy=require-corp"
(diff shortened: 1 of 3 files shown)


NickHu:
So there's a built-in one in firebase tools, which provides stuff like emulating the backend DB and things like that. My intention was that we would eventually use that, but it's maybe a bit a ways off still.

calintat:
The makefile is malformed btw:

```
[cargo-make] ERROR - Unable to parse external file: "./Makefile.toml", expected a right bracket, found a string at line 68 column 3
```

regular-citizen:
> The makefile is malformed btw:

Thanks for reporting this, fixed.

> So there's a built-in one in firebase tools, which provides stuff like emulating the backend DB and things like that. My intention was that we would eventually use that, but it's maybe a bit a ways off still.

Oh cool I didn't know this. The main thing is that right now devserver is broken on my machine. I think the best course of action would be for me to merge this PR unless there is strong opposition, and then you can open a new one in the future with that changes. I expect it will mostly look like the diff in this PR, adapted to the firebase webserver ofc.

calintat:
The makefile doesn't work for me :(

regular-citizen:
Do you have sfz installed?

calintat:
> Do you have sfz installed?

Yes!

regular-citizen:
Works on my machine, I only have a cargo watch error but the webserver spawns fine. Can you send me the error you're seeing?

calintat:
I'm not getting any errors. It says "build done" but when I go to localhost:5000, I see nothing.

regular-citizen:
Oh yeah sometimes that happens, try to do a force refresh.

calintat:
I've tried that numerous times, still doesn't work.

Issue #198, opened 2021-10-08
Signature suspension


vinnylarouge:
for a specific use case I have, my generators are all 2-cells on 1-cells of a single 0-cell (planar), but I would like to define a special 1-cell X that braids with the other 1-cells. Since it was less effort than specifying that many braiding rules, I promoted all my 2-cells to 3-cells and 1-cells to 2-cells.

would It be possible either to shortcut braiding equations, or to systematically promote/demote collections of cells to give/remove symmetries?

vinnylarouge:
(i might have to seek the symmetric closure of the dependency relation to pick out the legal subcollection to suspend)

calintat:
I think this is an interesting. We should at least provide an option to suspend the entire signature. Maybe also a subset of generators (though this might be trickier).

regular-citizen:
@mlharaujo also requested this feature (#788), here is his take on this:
> It would be useful to be able to take a signature and "suspend" it. The suspended signature would have either one or two 0-cells (this could be a user choice). Then it would have a (k+1)-cell for each k-cell in the original signature, with the obvious sources and targets. The 1-cells in the suspended signature would all have the same source and the same target.

regular-citizen:
This feature was requested by Hei Li as well.

Issue #197, opened 2021-10-08
more folder capabilities


vinnylarouge:
it would be useful to have the ability to make a duplicate of a folder and all contained cells.

jamievicary:
Interesting, why would you want to do this?

calintat:
This is related to #523. I think it is a good idea, we should be able to duplicate folders as well as individual generators.

regular-citizen:
This feature was requested by Hei Li as well.

PR #1022, opened 2023-02-18
Signature suspension (#198)

regular-citizen:
This PR implements signature suspension, as detailed in issue #198. In the UI, this has been added as a button in the debug panel.

It is non-trivial to get right because of the labelling logic, which is currently ignored. Do not use it on actual proofs until this is fixed.

My current implementation seems to be producing malformed rewrite which do not compose, so the inductive step of the suspension algorithm is definitely broken.


regular-citizen:
This seems to have fixed the panics. @calintat if you can figure out labels we can merge.

jamievicary:
This looks fantastic Chiara. I looked at this via the link above but couldn't see any UI mechanic. What is your plan there. A button at the top of the signature to avoid cluttering the left hand menu?

By the way suspension in general makes sense more locally, it doesn't have to be for the entire signature.

regular-citizen:
The current UI puts it as a button in the UI panel, which is not brilliant.

I have not thought too much about UI, generally suspending things seems like something that gets done rather infrequently, but if we want to support local suspensions (e.g. of folders), then some UI is really needed.

The local suspensions are likely going to be a messy thing to implement: if we suspend a generator, then we must ensure that every diagram in which that generator has ended up gets suspended as well. We could embrace this and allow to do selections of generators in the signature (with something like CTRL+CLICK) and then suspend them and their dependencies. The ability to select multiple generators/folders could then be recycled to implement a feature to identify generators (#523).

I definitely also want to support suspension with only one base generator, which is super quick  to implement.

jamievicary:
I looked at the preview URL but I can't see a button in the menu bar for this.

regular-citizen:
It's in the debug panel.
image

jamievicary:
Thanks this is interesting, I have now played with it a bit.

I was actually imagining a different notion of suspension, where everything lives over a single new object. This would turn monoidal category data into braided monoidal category data, braided monoidal into symmetric monoidal, etc. This is the notion of suspension that I am familiar with from this paper: https://arxiv.org/abs/q-alg/9503002

Can you explain a bit about this notion of suspension you've implemented? And let's discuss the different ways it could act, and what we'd use it for.

regular-citizen:
Small note to make is that we could get away with just doing the UI for the more general suspension (with two fresh 0-cells) if we had generator identifications (#197), as we could then combine the two new generators into a unique one.

jamievicary:
Hi Chiara, I still don't think that gives the procedure I was describing. If we start with an object of a monoidal category, this is an object X with an endomorphism f:X->X. After "suspending" this in Baez's sense, we should have f:id(X) -> id(X). But this is not what either of your suspension operations gives.

regular-citizen:
Oh okay I now understand, they are indeed different constructions, I stand corrected.

The problem I see with implementing the Baez-style suspension is that it fails globularity for non-endomorphisms. This means that the UI for it should be more granular, which requires a bit of discussion. But outside that it should be relatively straightforward to implement (without labels), and I can take a stab at it at some point.

Should I close this PR?

jamievicary:
Let's chat later this week.

calintat:
Ah interesting! I hadn't realised there is another notion of suspension.

The one we implemented is the standard notion of suspension from topology (https://ncatlab.org/nlab/show/suspension). This is also described in Thibaut's thesis for CaTT: given a context $\Gamma$, its suspension $\Sigma \Gamma$ is obtained by adding two fresh variables $s, t : *$ and replacing every $x : *$ with $x : s \to t$.

I suggest we call this *path suspension*, and the variant where $s = t$ can be called *loop suspension*.

Now John Baez's notion of suspension is a functor of $k$-tuply monoidal $n$-categories to $(k + 1)$-tuply monoidal $n$-categories. In terms of signatures, this can be implemented as follows. Given a signature $\Gamma$, there are two cases:
* If $\Gamma$ has more than one 0-generator, it's the same process as loop suspension.
* If $\Gamma$ has a unique 0-generator, say $x$, and no $j$-generators for $0 < j < k$, then replace every occurrence of $x$ with $\mathrm{id}(x)$.

I suggest we call this *abelianization*. This can be used to convert a set into a monoid into a commutative monoid, or a category into monoidal category into a braided monoidal category into a symmetric monoidal category, and so on.

jamievicary:
Thanks for that nice analysis Calin.

A subtlety is that we want the strict rather than weak identity $\text{id}(x)$ of course.

calintat (commented):
@regular-citizen left some comments, feel free to take a look whenever you want.

regular-citizen:
Will implement other ideas discussed in a future cleanup commit.

calintat:
@regular-citizen I realised we can combine the `suspend` and `abelianize` functions. I think it makes the code nicer.

regular-citizen:
> @regular-citizen I realised we can combine the `suspend` and `abelianize` functions. I think it makes the code nicer.
> 
> Basically the only difference between the two is whether you suspend over fresh generators or an existing generator.

This is great! Thanks for implementing this patch, it does make things much neater.

calintat:
Here is another idea: we can make suspension a button on a generator. You can abelianize by clicking suspend on the unique 0-cell, and you can do reduced suspension by first creating a fresh generator and then clicking suspend on that. The only downside is there would be no way to do standard suspension.

regular-citizen:
> Here is another idea: we can make suspension a button on a generator.

I don't think this is optimal because we would like to support standard suspension. However we agreed today with Jamie to implement the generator identification feature in a similar way. It would involve a button that identifies a generator with its predecessor, if the diagrams have compatible boundaries.

I definitely want to hide the suspension buttons from view if the signature is empty though.

calintat:
How about we have two buttons (suspend source and suspend target) and make them work like set boundary?

regular-citizen:
> How about we have two buttons (suspend source and suspend target) and make them work like set boundary?

I'm not sure I understand this.

calintat:
As in you press "suspend source" on a 0-generator and it remembers your choice and then when you press "suspend target" on either the same generator or a different, it does the suspension.

regular-citizen:
I'm not sure I like this too much but we would do something similar for generator identifications, so maybe it's worth considering.

homotopy-model/src/proof.rs
```diff
@@ -728,22 +725,41 @@ impl ProofState {
...
+            .filter(|g| g.dimension == 0)
+            .count();
+        let (source, target) = match (genz_count, kind) {
+            // If signature is empty quit
+            (0, _) => {
+                return false;
+            }
+            // If there is not a unique 0-cell we perform a reduced suspension
+            (1, SuspensionKind::Abelian) => {
+                let base = *self.signature.generator_iter().next().unwrap();
```
calintat, line 746: This might not be dimension 0 (if you reordered the generators) which causes a panic.

regular-citizen: Thanks for finding this! Fixed

regular-citizen:
We discussed this with @calintat, the UI should be a single "suspension" button on 0-cells. To do a reduced suspension, we would add a fresh generator, and do suspension on that, and abelianization could be done by suspending an existing generator.

Standard suspension would not be supported due to its failure modes.

The existing algorithm support all of this, so this can be implemented as soon as labels have been figured out.

calintat:
> Standard suspension would not be supported due to its failure modes.

We could support that as an action on the whole signature (same as it's already implemented).

Issue #522, opened 2022-06-20
Merge generators


calintat:
Select two generators (with identical boundaries) and make them the same.

regular-citizen:
We discussed this with @calintat today, the UI should be a "strictify" button on $(n+1)$-cells, which identifies the source and target of the cell and replaces it with an identity. This would only be defined for cells with atomic source and targets.

PR #1031, opened 2023-03-03
Generator identifications (#522)

regular-citizen:
This PR implements generator identifications. The UI is given by a process of strictification: to identify two generators, one adds an algebraic cell between them and then strictifies that away to an identity.


homotopy-core/src/diagram.rs
```diff
@@ -220,6 +220,38 @@ impl Diagram {
             Self::DiagramN(d) => d.suspend(s, t),
         }
     }
+
+    pub fn atomic_generator(&self) -> Option<Generator> {
+        match self {
+            Self::Diagram0(d) => Some(d.generator),
+            Self::DiagramN(d) if d.cospans().is_empty() => d.source().atomic_generator(),
```
calintat, line 227: I'm not sure an identity diagram should be atomic.

homotopy-core/src/diagram.rs
```diff
@@ -220,6 +220,38 @@ impl Diagram {
...
+                .rewrite_forward(&d.cospans()[0].forward)
+                .unwrap()
+                .atomic_generator(),
+            Self::DiagramN(_) => None,
+        }
+    }
+
+    pub fn atomic_boundaries(&self) -> Option<(Generator, Generator)> {
+        if let Self::DiagramN(d) = self {
+            match (d.source().atomic_generator(), d.target().atomic_generator()) {
```
calintat, line 239: You might want to use [zip](https://doc.rust-lang.org/std/option/enum.Option.html#method.zip).

homotopy-core/src/diagram.rs
```diff
@@ -405,6 +446,17 @@ impl DiagramN {
...
 
+    #[must_use]
+    pub fn replace(&self, s: Generator, t: Generator, a: Generator) -> Self {
+        let source = self.source().replace(s, t, a);
+        let cospans: Vec<_> = self
+            .cospans()
+            .iter()
+            .map(|c| c.map(|r| r.replace(s, t, a)))
+            .collect();
+        Self::new(source, cospans)
```
calintat, line 457: I wonder if we should have a `map` function on `DiagramN` since this pattern is becoming quite common.

homotopy-model/src/proof.rs
```diff
@@ -391,6 +397,45 @@ impl ProofState {
...
+    /// Returns an error if generator has dimension zero,
+    /// the generator is not in the signature or the boundaries are not atomic
+    fn strictify(&mut self, generator: Generator) -> Result<bool, ProofError> {
+        let info = self
+            .signature
+            .generator_info(generator)
+            .ok_or(ProofError::UnknownGeneratorSelected)?;
+
+        let Some((source, target)) = info.diagram.atomic_boundaries() else {
+            return Ok(false);
```
calintat, line 411: This should be an error type.

homotopy-model/src/proof.rs
```diff
@@ -391,6 +397,45 @@ impl ProofState {
...
+        };
+        if source.dimension != target.dimension || source == target {
```
calintat, line 413: I'm confused, when can the dimensions ever be different?

homotopy-core/src/diagram.rs
```diff
@@ -291,6 +323,15 @@ impl Diagram0 {
         }
     }
 
+    #[must_use]
+    pub fn replace(&self, s: Generator, t: Generator, a: Generator) -> Self {
+        if self.generator == s || self.generator == a {
+            Self::new(t, self.orientation)
```
calintat, line 329: I sort of imagined we would replace s and t with a (but with the dimension - 1).

homotopy-web/src/app/signature/item.rs
```diff
@@ -814,6 +814,11 @@ impl ItemView {
                 <button onclick={ctx.props().dispatch.reform(move |_| Action::Suspend(generator, generator))}>{"Suspend"}</button>
             </div>
         };
+        let strictification_button = html! {
```
calintat, line 817: I think we should hide the button if it is invalid.

homotopy-model/src/proof.rs
```diff
@@ -165,6 +169,7 @@ impl Action {
             Self::ClearWorkspace => proof.workspace.is_some(),
             Self::ClearBoundary => proof.boundary.is_some(),
             Self::SelectGenerator(_) => true,
+            Self::Strictify(g) => g.dimension > 0,
```
calintat, line 172: This should also return false if the source = target (basically the rule is `Action::is_valid` should return false whenever the action handler returns false).

homotopy-model/src/proof/signature.rs
```diff
@@ -99,6 +99,29 @@ impl Signature {
             .map_or(0, |id| id + 1)
     }
 
+    #[must_use]
+    pub fn filter_map<F>(&self, f: F) -> Signature
```
calintat, line 103: Could this be moved to homotopy_common? I.e. can this be done for any tree not just signatures?

homotopy-web/src/app/signature/item.rs
calintat: Actually on second thoughts, checking if it is valid can be expensive (as it involves checking if the source and target are atomic), so maybe it's not the best idea.

calintat:
Thanks Chiara! Hope my comments make sense (I am slightly drunk).

homotopy-model/src/proof.rs
regular-citizen: They can't. I was wrong.

homotopy-core/src/diagram.rs
regular-citizen: Yeah that would make sense, but it is a massive pain to implement. I think the current version is fine.

homotopy-web/src/app/signature/item.rs
regular-citizen: Agreed it's hard to check for it within the signature drawers.

regular-citizen:
Thanks for the feedback. Implemented as much as possible

PR #1041, opened 2023-03-11
CSS changes

regular-citizen:
Minor CSS cleanups, should be barely noticeable but make the website more reactive by removing some accumulated hacks.


regular-citizen:
Apparently causes breaking changes. Do not merge until that is addressed.

PR #1049, opened 2023-03-15
Generator identifications

regular-citizen:
This PR introduces new UI for generator identifications.


regular-citizen:
Thanks for the patch. I'll see if we can merge tomorrow.

regular-citizen:
@calintat you patch seems to make a test fail. I think you framing removal logic is wrong. I'll patch it later, but it's quite worrying that CI did not pick it up.

calintat:
Thanks, I just fixed it.

homotopy-core/src/rewrite.rs
```diff
@@ -377,7 +377,7 @@ impl Rewrite0 {
                 target.replace(from, to),
                 label
                     .as_ref()
-                    .filter(|_| oriented && (target.generator == from || target.generator == to))
+                    .filter(|_| !oriented || (target.generator != to && target.generator != from))
```
calintat, line 380: Ooops! Thanks for fixing this!

Issue #86, opened 2021-05-25
Interrupt workflow to add a new generator


NickHu:
Frequently I have a workspace in which I am in the middle of trying to construct some proof, only to make the realisation that I need to add some lemma to continue the proof. Currently, I have to clear my workspace (thus abandoning my current proof attempt) to add this lemma. It would be good if we had a workflow to interrupt mid-proof to add a lemma or something like that.

jamievicary:
You can do it using the "theorem" mechanism to store your proof, then
retrieve it later by navigating to the appropriate slice of the theorem and
selecting "restrict".

But yes, I agree it would sometimes be convenient to be able to stash
diagrams more directly. Maybe a diagram drawer would make sense.

H-Yulong:
You also stash diagrams with source/target preview now

jamievicary:
Very good point that's interesting.

jamievicary:
It's sort of awkward because it seems like a "hack" using this to stash diagrams.

calintat:
Core functionality implemented by 12773ed917f3e0aa00f0ee1209a679068333ef9b, with the UI to be done in the future.

PR #1051, opened 2023-03-20
Stash diagrams

calintat:
(I wrote this a while ago so I thought I'd push it in case anyone wants to take it further).

Basic implementation of #86 in a similar style to `git stash`. It has no UI right now, just keybindings (`+` to stash, `-` to pop).

### DIFF
--- a/homotopy-model/src/proof.rs
+++ b/homotopy-model/src/proof.rs
@@ -78,7 +78,7 @@ pub struct ProofState {
     pub workspace: Option<Workspace>,
     pub metadata: Metadata,
     pub boundary: Option<SelectedBoundary>,
-    pub stack: Vector<Workspace>,
+    pub stash: Vector<Workspace>,
 }
 
 #[derive(Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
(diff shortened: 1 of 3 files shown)


regular-citizen:
Thanks for coding this, I like this a lot.

Two things:
- Stashing a diagram, then deleting a generator used in it, results in a panic when unstashing it.
- You should not be able to unstash a diagram if the current workspace diagram is not empty in my opinion.

Should we support keeping the current diagram while stashing it?

calintat:
I think it's fine if unstashing works even if the workspace is not empty, because selecting a generator also works when the workspace is non-empty. Plus you can always undo if you make a mistake.

calintat:
I think it's weird if stashing didn't clear the workspace.

regular-citizen:
> I think it's weird if stashing didn't clear the workspace.

Yeah I have realised that we can already keep the current diagram by saving it as source/target, then selecting it again and stashing it. So the current flow works.

calintat:
Once we figure out the UI, we can also have an action that recovers the top stashed diagram without removing it from the stack (similar to `RecoverBoundary`).

PR #1234, opened 2023-09-22
Initial account support

thud:
This PR adds initial account support with Firebase to allow saving projects to the cloud.

#### Important
- [x]  Create a full backup of existing Firestore database from legacy homotopy-io. [see [firebase docs](https://firebase.google.com/docs/firestore/manage-data/export-import)]
- [x] Migrate legacy projects over to homotopy-rs.
- [x] Catch URLs client-side to allow for project permalinks etc.

#### Less important:
- [ ] Restyle project UI in accounts drawer.
  - [ ] Add delete button for projects (functionality already built).
  - [ ] Make publishing UI clearer.
- [ ] Alerts for user when saving to avoid unwanted overwrites.
- [ ] Autosave functionality (or always visible save button).
- [ ] Catch and display errors properly in Javascript code (CFs and client-side).
- [ ] Implement a "gallery" view to see other users' recent projects.
- [ ] Rewrite storage rules to be as strict as possible (but without breaking legacy homotopy-io).
- [ ] Document Cloud Functions and client code for maintenance and clarity.


NickHu:
Hi @thud, trying to get round to integrating this now.
I reviewed the code, and it mostly looks good, I am just confused about one aspect.
One of the major issues we are trying to solve is having some kind of atomicity with respect to both Firestore and Firebase Storage, so that you can't have invalid database entries or dangling storage objects etc.
This is essentially why we are using Cloud Functions.

I guess the nonce stuff you have is in order to guard against data races of this form?

In this function https://github.com/homotopy-io/homotopy-rs/blob/b386995a1d23be6ad87ea6ceb7b38f8a760c98a6/homotopy-web/src/app/account/account_script.js#L91-L151
which is called from the client, first a Cloud Function call is made to generate some db entry on Firestore, and then the client handles the upload directly using Firebase Storage - but why not instead just have the blob & metadata uploaded in a single HTTP request to a Cloud Function endpoint which is designed to handle it? One of the problems with the previous design was in making multiple write calls to different Firebase services, because if the program crashed in between you would end up in an inconsistent state.

thud:
Hi @NickHu. Yeah can't remember exactly why at the moment but will have a think over the next few days.

I might be wrong but as I recall, I think cloud functions can't accept blobs as inputs so instead you have to write blobs directly to Storage and use event triggers to detect the write and update Firestore. I tried this approach with metadata stored in the `custom_metadata` of the blob upload. I'm pretty sure it worked ok but I ran into issues somewhere down the line with data desynchronisation or something similar.

I also implemented full versioning for projects on the backend (with individual metadata for each version etc) and I think this might have necessitated the `requestBlobUpload` approach. Later, Jamie told me to revert all that so I probably just chopped off the versioning and left the rest as is to save time.

The nonce stuff was to ensure that a certain set of metadata exactly corresponded to a certain blob but I'm not sure that I even implemented the checks so maybe that's useless as it stands anyway (not sure).

The solution I wrote in the end is probably overly complex and subtly non-atomic anyway but it felt better to me at the time. Obviously please feel free to rip it out and reimplement as desired. Hopefully this is helpful.

NickHu:
Will merge this for now with a big warning.

I think the way to do this properly is to have the client only write to Firebase Storage, and have a Cloud Function keep the metadata bookkeeping updated in Firestore - to revisit later.

PR #1462, opened 2024-10-14
Fix manim

robin-carlier:
Currently, the `.py` files exported by (the beta version of) homotopy.io contain syntax errors and accesses to nonexisting elements of dictionnaries. I think this should fix it. I am unsure about the saturation to put at intersections though.


regular-citizen:
Thanks for the patch! Really helpful. Will merge

